PR #4, opened 2015-08-16
Add separate module for annotations

hvisser:
This adds a separate module for the annotations so that the processor no longer depends on the Android library project. This is not allowed in Android builds and gives a warning. Updated `android-apt` to 1.7 and updated the sample project dependency to `provided`. Normally `apt` should work as well, but I guess that's a bug in `android-apt` that I have to fix ;)

I'm not 100% sure if publishing to Maven works correctly with the extra module addition.


hotchemi:
Wow great! This is actually I've been thinking.
But I'd like to ask you an one point,

When it comes to separate to annotations and "core" module, user have to write 3 dependencies definition in their build.gradle, I think it's not major style, I mean some famous library(like butterknife, auto-parcel, deeplinkdispatch) give two dependency, core and "processor".

How do you think about of it?

hotchemi:
There is no problem in changes.

hvisser:
You can have two dependencies if you have normal jar projects, like `auto-parcel` for example. If you have an `android-library` project, the processor cannot have a dependency on that since `javac` does not know about that on the class path. Butterknife only has one jar with everything packed into that, but that's not as nice, since the processor will end up in your apk file, unless you use ProGuard.

It's kind of annoying, but there's basically a world for the annotation processor at compile time and one for the Android code at run time.

hotchemi:
I see. But actually I wanna check everything goes well before merge it to master branch. So could you send a pull request to develop branch again? Anyway, I really appreciate for you contribution!

hotchemi:
So I'm gonna close the issue.

hvisser:
Ehm OK, the develop branch did not exist when sent you the PR, so this should still apply to that branch.

hotchemi:
Yeah, I'm really sorry for that.

PR #13, opened 2015-08-26
#12: Add Permission Denied handling

### LINKED ISSUE #12: Add Permission Denied handling
At the moment, PermissionsDispatcher's generated code for `onRequestPermissionsResult` verifies the grant results and invokes the positive callback annotated with `@NeedsPermission(s)` in case the permissions were granted. How about we add support for negative callbacks, accompanied by an annotation named `@DeniedPermission(s)`, invoked when the user denies a requested Runtime Permission? This would be different from `@ShowsRationale(s)`, which triggers an action at the time the permission is already being asked for.

Here's an example I threw together:

```java
@RuntimePermissions
public class MyActivity extends AppCompatActivity {
	
	// The usual constants and members (...)

	@Override public void onCreate(Bundle savedInstanceState) {
		// Some other setup code (...)

		// Add a click listener to some button, initializing a "dangerous action" requiring a Runtime Permission
		someButton.setOnClickListener(v -> MyActivityPermissionsDispatcher.setupCameraWithCheck(this));
	}

	@NeedsPermission(Manifest.permission.CAMERA) void setupCamera() {
		// Perform "dangerous action", now that permission is granted
	}

	@ShowsRationale(Manifest.permission.CAMERA) void showRationaleForCamera() {
		// Show a small message accompanying the permission dialog
	}

	@DeniedPermission(Manifest.permission.CAMERA) void onCameraPermissionDenied() {
		// React to the user denying the permission, maybe show a dialog?
	}

	@Override public void onRequestPermissionsResult(int requestCode, @NonNull String[] permissions, @NonNull int[] grantResults) {
		// Propagate the result
		MyActivityPermissionsDispatcher.onRequestPermissionsResult(this, requestCode, grantResults);
	}
```

```java
public final class MyActivityPermissionsDispatcher {
	private static final int REQUEST_SETUPCAMERA = 0;

	private static final String[] PERMISSION_SETUPCAMERA = new String[]{"android.permission.CAMERA"};

	private MyActivityPermissionsDispatcher() {
	}

	public static void setupCameraWithCheck(MyActivity target) {
		if (PermissionUtils.hasSelfPermissions(target, PERMISSION_SETUPCAMERA)) {
      		target.setupCamera();
    	} else {
	    	if (PermissionUtils.shouldShowRequestPermissionRationale(target, PERMISSION_SETUPCAMERA)) {
	    		target.showRationaleForCamera();
	    	}
      		ActivityCompat.requestPermissions(target, PERMISSION_SETUPCAMERA, REQUEST_SETUPCAMERA);
    	}
  	}

	public static void onRequestPermissionsResult(MyActivity target, int requestCode, int[] grantResults) {
    	switch (requestCode) {
      		case REQUEST_SETUPCAMERA:
      			if (PermissionUtils.verifyPermissions(grantResults)) {
        			target.setupCamera();
      			} else {
      				// Invoke the "permission denied" callback
      				target.onCameraPermissionDenied();
      			}
      			break;
		default:
      			break;
    	}
  	}
}
```

aurae:
Here is an implementation for @DeniedPermission annotations. I have seen that you are currently re-writing the processor in Kotlin, however I chose to extend the Java implementation to avoid conflicts.

* Added new annotations @DeniedPermission & @DeniedPermissions, which can be used to react to denied requests for permissions
* Extended annotation processor to include the corresponding callbacks
* Added tests for the new annotations

This fixes #12


hotchemi:
First of all, thx for your great work. I'm gonna check it.

permissionsdispatcher-processor/src/main/java/permissions/dispatcher/processor/JavaFileBuilder.java
```diff
@@ -133,7 +147,7 @@ static JavaFile createJavaFile(RuntimePermissionsAnnotatedElement element) {
                 .addMethod(createConstructor())
                 .addMethods(createNeedsPermissionMethods(element))
                 .addMethods(createNeedsPermissionsMethods(element))
-                .addMethod(createOnRequestPermissionsResult(element.getClassName(), permissionsMethods))
+                .addMethod(createOnRequestPermissionsResult(element.getClassName(), permissionsMethods, element))
```
hotchemi: Please change arguments to only 'element'. Because first and second are valued of element.

permissionsdispatcher-processor/src/main/java/permissions/dispatcher/processor/JavaFileBuilder.java
```diff
@@ -74,12 +74,26 @@ private static MethodSpec createOnRequestPermissionsResult(ClassName target, Lis
...
-        for (ExecutableElement method : methods) {
+        for (ExecutableElement method : permissionMethods) {
             String methodName = method.getSimpleName().toString();
             methodBuilder
                     .addCode("case $N:\n", getRequestCodeFieldName(methodName))
                     .beginControlFlow("if ($T.verifyPermissions(grantResults))", PERMISSION_UTILS)
-                    .addStatement("target.$N()", methodName)
+                    .addStatement("target.$N()", methodName);
+            NeedsPermission needsPermission = method.getAnnotation(NeedsPermission.class);
+            String needsPermissionValue = needsPermission != null ? needsPermission.value() : "";
```
hotchemi: You don't have to handle null case because we get only annotated methods.

permissionsdispatcher-processor/src/main/java/permissions/dispatcher/processor/JavaFileBuilder.java
```diff
@@ -74,12 +74,26 @@ private static MethodSpec createOnRequestPermissionsResult(ClassName target, Lis
...
+            ExecutableElement deniedPermission = clazz.getDeniedPermissionFromValue(needsPermissionValue);
+            if (deniedPermission == null) {
+                NeedsPermissions needsPermissions = method.getAnnotation(NeedsPermissions.class);
+                String[] needsPermissionsValue = needsPermissions != null ? needsPermissions.value() : new String[0];
```
hotchemi: You don't have to handle null case because you get only annotated methods.

permissionsdispatcher-processor/src/main/java/permissions/dispatcher/processor/JavaFileBuilder.java
```diff
@@ -74,12 +74,26 @@ private static MethodSpec createOnRequestPermissionsResult(ClassName target, Lis
...
+        for (ExecutableElement method : permissionMethods) {
             String methodName = method.getSimpleName().toString();
             methodBuilder
                     .addCode("case $N:\n", getRequestCodeFieldName(methodName))
                     .beginControlFlow("if ($T.verifyPermissions(grantResults))", PERMISSION_UTILS)
-                    .addStatement("target.$N()", methodName)
+                    .addStatement("target.$N()", methodName);
+            NeedsPermission needsPermission = method.getAnnotation(NeedsPermission.class);
+            String needsPermissionValue = needsPermission != null ? needsPermission.value() : "";
+            ExecutableElement deniedPermission = clazz.getDeniedPermissionFromValue(needsPermissionValue);
```
hotchemi: I'd like to pull out codes which try to get 'deniedPermission' to single static method like 'getDeniedPermission' or like that. Because it simplifies a logic flow.

permissionsdispatcher-processor/src/test/java/permissions/dispatcher/processor/data/Source.java
```diff
@@ -626,4 +626,124 @@
                 "}"};
     }
 
+        public static class DeniedPermission {
```
hotchemi: Could you fit naming to method name?

permissionsdispatcher-processor/src/test/java/permissions/dispatcher/processor/data/Source.java
```diff
@@ -626,4 +626,124 @@
...
+                        "      break;",
+                        "      default:",
+                        "      break;",
+                        "    }",
+                        "  }",
+                        "}"
+                };
+        }
+
+        public static class DeniedPermissionNotMatching {
```
hotchemi: Could you fit naming to method name?

hotchemi:
About processor test, could you add these cases? The change will be perfect!

- you have one `DeniedPermissions` which fits to `NeedsPermissions` (valid)
- you have one  `DeniedPermission` and there is no fitted `NeedsPermissions` (valid)
- you have duplicated value in `DeniedPermission` and `DeniedPermissions` (invalid)
- you have private methods annotated with `DeniedPermission` and `DeniedPermissions` (valid)

hotchemi:
@aurae Just finished code review. I really appreciated for your contribution :+1:

permissionsdispatcher-processor/src/main/java/permissions/dispatcher/processor/JavaFileBuilder.java
aurae: That makes total sense, I will change that

permissionsdispatcher-processor/src/main/java/permissions/dispatcher/processor/JavaFileBuilder.java
aurae: I ran into an NPE while working on the PR earlier, though. I think that might be related to the library providing two Annotations expressing the need for permissions, @NeedsPermission and @NeedsPermissions. I will look into this again

permissionsdispatcher-processor/src/main/java/permissions/dispatcher/processor/JavaFileBuilder.java
aurae: Gotcha!

permissionsdispatcher-processor/src/test/java/permissions/dispatcher/processor/data/Source.java
aurae: Will do once I'm adding the other test cases

aurae:
> you have private methods annotated with DeniedPermission and DeniedPermissions (valid)

Wouldn't that be an invalid case because of the `private` modifier?
Thank you for the code review! I will update the PR shortly.

permissionsdispatcher-processor/src/main/java/permissions/dispatcher/processor/JavaFileBuilder.java
hotchemi: @aurae Sorry my point was wrong, permissionMethods contains NeedsPermission and NeedsPermissions.
In this case, you can use `Utils. getValueFromAnnotation`. Please check it :)

permissionsdispatcher-processor/src/main/java/permissions/dispatcher/processor/JavaFileBuilder.java
hotchemi: https://github.com/hotchemi/PermissionsDispatcher/pull/13/files#r38018896

hotchemi:
1. Sorry I mean "valid" is "not throwing RuntimeException". So you can test when fitted NeedsPermissions is not in the class DeniedPermission annotated class never be called.
2. I mean I'd like to check Exception will be thrown or not, like [that](https://github.com/hotchemi/PermissionsDispatcher/blob/d2684c3a6794f90164b8b69049c54836885cf240/permissionsdispatcher-processor/src/test/java/permissions/dispatcher/processor/PermissionsProcessorTest.java#L156)

aurae:
1. Do we want to allow that? I would argue that a "Denied" callback without a matching "Needs" method doesn't provide much utility, and might even cause confusion for a library's user (i.e. "why is my callback not invoked?"). For now, I will implement the test case like you said, though!

2. Okay, so we are talking about the same thing, then :) just wanted to make sure.

hotchemi:
1. That is make sense. But there is no implementation right? May be you can check in RuntimePermissionsAnnotatedElement's constructor but it is not necessary stuff but better.

aurae:
@hotchemi I have updated the PR according to our discussion! :+1:

hotchemi:
@aurae really thx! I'm gonna update readme and some stuff and release 1.2.0!

PR #47, opened 2015-11-27
Fix travis build

esmasui:


hotchemi:
Thx!

hotchemi:
Oh something is still wrong we'll check it.

Issue #51, opened 2015-12-03
READ_EXTERNAL_STORAGE broken on API 15 device


robinstark:
This will also apply to other permissions that were only introduced in newer Android versions than the app is running on.

A method in my app requires READ_EXTERNAL_STORAGE permission, but this permission was only introduced in API 16. When running the generated ...WithCheck() method on an API 15 device, PermissionsUtils.hasSelfPermissions will never return true (since the READ_EXTERNAL_STORAGE permission doesn't exist), and so the target method will never get called - even though the app will run fine without that permission. PermissionsDispatcher acts as if the permission has been denied.

I'm working around it with
```java
if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
    MyActivityPermissionsDispatcher.doThingWithCheck(this);
} else {
    doThing();
}
```

but that feels broken and it would be much cleaner if I didn't have to.

aurae:
This behaviour occurs because of an architectural decision in the Android framework, where non-existent permissions are never granted by design. I see the point about performing conditional checks on certain permissions that were added later in the Android lifecycle. This would most likely be enforceable in `PermissionsUtils` using a static mapping of those permissions to their minimum SDK level.

Mark Murphy published [a great article](https://commonsware.com/blog/2015/11/09/you-cannot-hold-nonexistent-permissions.html) on the exact same issue on his blog recently.

robinstark:
Awesome, thanks. I'll look forward to 2.0.1. My workaround feels like a bit too much of a blunt instrument, applying checks specifically to the permissions and devices that need them is a much better idea.

aurae:
This is what my sketch looks like at the moment:

```java
public final class PermissionUtils {

    // Map of dangerous permissions introduced in later framework versions.
    // Used to conditionally bypass permission-grant checks on older devices.
    private static final SimpleArrayMap<String, Integer> MIN_SDK_PERMISSIONS;
    static {
        MIN_SDK_PERMISSIONS = new SimpleArrayMap<>(6);
        MIN_SDK_PERMISSIONS.put("com.android.voicemail.permission.ADD_VOICEMAIL", ICE_CREAM_SANDWICH);
        MIN_SDK_PERMISSIONS.put("android.permission.BODY_SENSORS", KITKAT_WATCH);
        MIN_SDK_PERMISSIONS.put("android.permission.READ_CALL_LOG", JELLY_BEAN);
[...]

PR #60, opened 2015-12-10
fix hotchemi/PermissionsDispatcher#51 based on @aurae sketch code

sibeliusseraphini:


aurae:
Lgtm, aside from the hard-coded values that I included in the code, not remembering that constants get resolved to Strings anyways.

sibeliusseraphini:
@aurae I've tried using only the put, but didn't work either, only using the hardcoded strings

hotchemi:
Could you add the test code?

library/src/main/android/permissions/dispatcher/PermissionUtils.java
```diff
@@ -9,13 +10,30 @@
...
+import static android.os.Build.VERSION_CODES.KITKAT_WATCH;
+import static android.os.Build.VERSION_CODES.JELLY_BEAN;
 import static android.support.v4.content.PermissionChecker.checkSelfPermission;
 
 public final class PermissionUtils {
 
+    // Map of dangerous permissions introduced in later framework versions.
+    // Used to conditionally bypass permission-hold checks on older devices.
+    private static final SimpleArrayMap<String, Integer> MIN_SDK_PERMISSIONS;
+    static {
```
hotchemi: How did you find these candidates? by hand?

aurae: Yes, I hand-picked these permissions using the Android docs for Manifest.permission](http://developer.android.com/reference/android/Manifest.permission.html), extracting all permissions that were listed as "dangerous" **and** added in a later API version.

aurae:
When using the constants, what didn't work for you? I'll verify this later today, but there shouldn't be any difference in behaviour, since those values get compiled into their String contents anyways.

sibeliusseraphini:
My device (android 4.0) didn't recognize Manifest.permission.READ_EXTERNAL_STORAGE, on monday I will try to add some tests using roboeletric to test it against differents versions of Android devices

aurae:
It would make sense for your 4.0 device to disregard `READ_EXTERNAL_STORAGE`, since that wasn't added to the platform until 4.1.

sibeliusseraphini:
If I ask READ_EXTERNAL_STORAGE permissions using Android M style, it will always denied it for my older device, when I use the hardcoded strings it works, but with Manifest.permission.READ_EXTERNAL_STORAGE it doesn't work

aurae:
I see... this is very odd indeed. I try to find time today to check this out myself. Thanks for your report!

aurae:
Okay, I just double-checked the code, and the usage of `Manifest.permission.*` constants is working as expected. There must've been something off about your code, since I am able to verify that the behaviour of API-level dependent permissions works without the hard-coded strings as well.

Using the extended implementation of `hasSelfPermissions()` and `MIN_SDK_PERMISSIONS` populated as such:
```java
private static final SimpleArrayMap<String, Integer> MIN_SDK_PERMISSIONS;
static {
	MIN_SDK_PERMISSIONS = new SimpleArrayMap<>(6);
	MIN_SDK_PERMISSIONS.put(Manifest.permission.ADD_VOICEMAIL, ICE_CREAM_SANDWICH);
	MIN_SDK_PERMISSIONS.put(Manifest.permission.BODY_SENSORS, KITKAT_WATCH);
	MIN_SDK_PERMISSIONS.put(Manifest.permission.READ_CALL_LOG, JELLY_BEAN);
	MIN_SDK_PERMISSIONS.put(Manifest.permission.READ_EXTERNAL_STORAGE, JELLY_BEAN);
	MIN_SDK_PERMISSIONS.put(Manifest.permission.USE_SIP, GINGERBREAD);
	MIN_SDK_PERMISSIONS.put(Manifest.permission.WRITE_CALL_LOG, JELLY_BEAN);
}
```

Given the following statement (in an Activity's `onCreate()`):
```java
Log.d("Check", "Permission Value: " + Manifest.permission.READ_EXTERNAL_STORAGE + "; do I have that? " + hasSelfPermissions(this, Manifest.permission.READ_EXTERNAL_STORAGE));
```

This prints the following on a device running 2.3.7 Gingerbread:

```
12-14 08:37:40.364 7410-7410/com.example D/Check: Permission Value: android.permission.READ_EXTERNAL_STORAGE; do I have that? true
```

Therefore, please update your PR to use the constants mentioned above, so that we can merge it into the codebase.

sibeliusseraphini:
Sorry, it was my fault, now it is working fine, thanks.

aurae:
Thanks!

hotchemi:
@sibeliusseraphini @aurae Could you add unit test?

hotchemi:
I think we should not merge pr without unit test...

aurae:
Efficiently unit testing API-level dependent code requires us to set up a whole new test environment, using Mockito for instance. I didn't find it *that* necessary to do for this case right away, since it's a fairly simple concept, however I do see your point. I probably can't get to this until after Christmas, though.

sibeliusseraphini:
I think we should run this test using [Instrumented Unit Tests](http://developer.android.com/training/testing/unit-testing/instrumented-unit-tests.html), because the tests should run on real devices with different API levels

aurae:
I have [added some unit tests for this](https://github.com/hotchemi/PermissionsDispatcher/blob/master/library/src/test/java/permissions/dispatcher/ApiLevelTestSuite.java).

PR #85, opened 2016-01-28
Pull request to help reduce writing multiple @OnShowRationale methods

aneophyte:
Added a `getPermissionName()` method in `PermissionRequest` interface. Users can now query for what permission the rationale is to be shown.

Please check and pull if good. Thanks!


aneophyte:
Two of the tests are failing, which I'm hoping are unrelated to the changes made here.
http://pastie.org/private/4qxmzqwho7vegiso8jtpa

shiraji:
We won't be able to merge pull request that fail test cases. Let me find the issue.

@hotchemi I think we should have `--stacktrace` option for circle ci. WDYT?

hotchemi:
@shiraji u mean travis ci? We can do that.

shiraji:
BTW, why the developer needs to know which permission rationale is shown? Could you give us a sample code that require to call this method? Because if the developer want to differentiate the permission like sample code

```java
    @OnShowRationale({Manifest.permission.READ_CONTACTS, Manifest.permission.WRITE_CONTACTS})
    void showRationaleForContact(PermissionRequest request) {
        showRationaleDialog(R.string.permission_contacts_rationale, request);
    }
```

Then, the developer should create methods for each permission.

```java
    @OnShowRationale(Manifest.permission.READ_CONTACTS)
    void showRationaleForContact(PermissionRequest request) {
        showRationaleDialog(R.string.permission_contacts_rationale, request);
    }

    @OnShowRationale(Manifest.permission.WRITE_CONTACTS)
    void showRationaleForContact2(PermissionRequest request) {
        showRationaleDialog(R.string.permission_contacts_rationale, request);
    }
```

Am I missing something?

shiraji:
Yea, travis ci. I was confused with other project.

processor/src/test/java/permissions/dispatcher/processor/data/Source.java
```diff
@@ -1574,9 +1598,6 @@ protected String getName() {
                     "        }",
                     "        @Override",
                     "        public void cancel() {",
-                    "            MyActivity target = weakTarget.get();",
-                    "            if (target == null) return;",
-                    "            target.cameraDenied();",
```
shiraji: Why did you remove this code? This makes tests fail.

processor/src/test/java/permissions/dispatcher/processor/data/Source.java
```diff
@@ -2392,9 +2425,6 @@ protected String getName() {
...
```
shiraji: This one, too.

aneophyte:
Sorry, my bad. I was trying to get the tests to run and might have deleted by accident. Were those sources for the tests made by hand? Must have been tedious.

By the way, the reason for this pull request is to be able to avoid writing `@OnShowRationale` multiple times. The user can:

`@OnShowRationale({Manifest.permission.READ_CONTACTS, Manifest.permission.WRITE_CONTACTS})
    void showRationaleForContact(PermissionRequest request) {
        showRationaleDialog(request.getPermissionName(), request);
        //or showRationaleDialog(request); even
    }`

So the PR can't be merged unless the checks are made to pass?

shiraji:
Well, it's up to other maintainers. They think this function is useful for this project, then yes. However, I personally don't see the point to have this function in order to avoid writing `@OnShouRationale`, if there is other use case for this, I might agree.

> So the PR can't be merged unless the checks are made to pass?

I don't think any of project want to merge pull request that fail test cases.

aneophyte:
I guess this is preference over value type thing. I wanted to map the rationale dialog logic in one place and came up with this. PR is already in place, hopefully the maintainers can merge with little to no issue. :P

hotchemi:
IMO, I'm not a huge fun of this idea. The reason is it constrains users to implement `getPermissionName` even if they don't want to do.
The solution is what @shiraji said. When you want to show rationale stuff by each permission, you should request permissions one by one.

hotchemi:
And when it comes to test source, yes it's so massive so we're trying to separate to each java file.

aneophyte:
The users don't even have to care about `getPermissionName()`, they get the option if they wish to go that route.

processor/src/main/kotlin/permissions/dispatcher/processor/impl/BaseProcessorUnit.kt
```diff
@@ -406,6 +406,14 @@ public abstract class BaseProcessorUnit : ProcessorUnit {
         }
         builder.addMethod(constructorBuilder.build())
 
+        // Add getPermissionName() override
+        val getPermissionNameMethod : MethodSpec.Builder = MethodSpec.methodBuilder("getPermissionName")
+                .addAnnotation(Override::class.java)
+                .addModifiers(Modifier.PUBLIC)
+                .returns(ClassName.get(String::class.java))
+                .addStatement("return \"${needsMethod.getAnnotation(NeedsPermission::class.java).value[0]}\"")
```
hotchemi: `value[0]` always returns only first element.

hotchemi:
Sorry, but this issue is still there, it always returns only first value.

https://github.com/hotchemi/PermissionsDispatcher/pull/85#discussion_r51228668

aneophyte:
Oops. Will get that fixed.

aneophyte:
This fixes that. Your call guys.

shiraji:
Now, how the developer knows which permission is to be shown? You return all of them.

aneophyte:
Just realized that it fails in another way. The processor seems to work with chunks of permission request rather than the individual permission. A lot more is to be done to achieve this and I underestimated the level of change needed. I'm closing this PR.

By the way, it doesn't return all of them. Just the array from `@NeedsPermission`.

Thanks for your time guys. Might open up another one next time.

hotchemi:
OK, Thanks for your contribution  @aneophyte .

PR #86, opened 2016-01-31
Common @OnShowRationale method - attempt no. 2

aneophyte:
This is another attempt at unifying `@OnShowRationale` methods. Here, the user will annotate a method without any parameters to receive calls for multiple permission request types.

Change is sort of drastic, touched some more files than before. Did this just for completion's sake, feel free to reject this pull request and close if this is not preferred. I understand the cost of maintenance that adding this might take, so many tests had to be added.


shiraji:
I think there is missing test cases. Multiple `NeedsPermission` and `OnShowRational`.
Like this one

```java
@NeedsPermission({Manifest.permission.CAMERA, Manifest.permission.READ_CONTACTS})
void needsPermission() {
}

@OnShowRational({Manifest.permission.CAMERA, Manifest.permission.READ_CONTACTS})
void onShowRational(PermissionRequest request) {
}
```

Maybe I misread this, but could you explain how the developer handles this case?

Expected generate source would be this

```java
private static final String[] PERMISSION_SHOWCAMERA = new String[] {\"android.permission.CAMERA\", \"android.permission.READ_CONTACTS\"};",
...

@Override
public String[] getPermissionNames() {
    return PERMISSION_SHOWCAMERA;
}
```

If the developer use `getPermissionNames()`, they get both `Manifest.permission.CAMERA, Manifest.permission.READ_CONTACTS`, right?

---

Could you do me a favor?

* Use non-master branch
* Make commits separately
* Append test cases, not add test cases middle of `source.java` (which is not rule in general but since the file is too big and most cases are similar, it's really hard to check diff.)

aneophyte:
Also, another case could be the mismatch of permissions. For example, if the user requests for camera and contacts permission in `@NeedsPermission` and forgets to include exactly that set of permission in `@OnShowRationale`, the `@OnShowRationale` annotated method will be disregarded by the processor. That means it would be as if no `@OnShowRationale` method is present. But that I think is out of the semantics, purely an error on the user's side and shouldn't be part of the test.

```java
public static final String[] PERMISSIONS = {Manifest.permission.CAMERA, Manifest.permission.READ_CONTACTS};

@NeedsPermission({Manifest.permission.CAMERA, Manifest.permission.READ_CONTACTS})
void needsPermission() {
}
@OnShowRational({Manifest.permission.CAMERA, Manifest.permission.READ_CONTACTS})
void onShowRational(PermissionRequest request) {
  showRationaleDialog(request);
}
void showRationaleDialog(PermissionRequest request) {
  if(Arrays.equals(PERMISSIONS, request.getPermissionNames()) {
    //do stuff
  }
}
```
That would be in the previous case. But actually it can be done in the following way.
```java
public static final String[] PERMISSIONS_CAMERA = {Manifest.permission.CAMERA};
public static final String[] PERMISSIONS_CONTACT = {Manifest.permission.READ_CONTACTS};

@NeedsPermission({Manifest.permission.CAMERA})
void needsPermissionCamera() {
}
@NeedsPermission({Manifest.permission.READ_CONTACTS})
void needsPermissionContact() {
}
@OnShowRationale
void onShowRationale(PermissionRequest request) {
  showRationaleDialog(request);
}
void showRationaleDialog(PermissionRequest request) {
  if(Arrays.equals(PERMISSIONS_CAMERA, request.getPermissionNames()) {
    //do stuff
  }
 if(Arrays.equals(PERMISSIONS_CONTACT, request.getPermissionNames()) {
    //do other stuff
 }
}
```
In all honesty, I just realized that Java doesn't allow Array constants to be passed as annotation parameters. Also, the more I do this, the more its worth is getting less.

Yes, certainly, will get what you've told done today.

hotchemi:
Sorry, we don't have a plan to support this change.

aneophyte:
Yep, understood. Too much to maintain for so much little.

PR #92, opened 2016-02-14
Cache get target sdk version.

hotchemi:
Just modify to cache targetSdkVersion to static variable.
This method might be called plural times, so it would improve perfomance.


hotchemi:
Ah I squashed master branch so I'll send it again.

PR #93, opened 2016-02-14
Modify to cache target sdk version.

hotchemi:
Just modify to cache targetSdkVersion to static variable.
This method might be called plural times, so it would improve perfomance.

### DIFF
--- a/library/src/main/android/permissions/dispatcher/PermissionUtils.java
+++ b/library/src/main/android/permissions/dispatcher/PermissionUtils.java
@@ -27,7 +27,7 @@ public final class PermissionUtils {
         MIN_SDK_PERMISSIONS.put("android.permission.WRITE_CALL_LOG", 16);
     }
 
-    private static int targetSdkVersion = -1;
+    private static volatile int targetSdkVersion = -1;
 
     private PermissionUtils() {
     }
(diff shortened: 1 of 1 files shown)


hotchemi:
Changed assignee to shiraji cuz aurae is on trip now.

aurae:
Why is `targetSdkVersion` an `Integer` rather than a primitive?

hotchemi:
I thought we can't check the variable is already set or not. But umm, we can check it as 0 or not right?

hotchemi:
Fixed.

hotchemi:
Umm, I have to choose double check pattern like Picasso...it's not clean code but we have to take care of performance...

hotchemi:
@shiraji thoughts?

shiraji:
Too little too late...but does it need to be `synchronized`? We can accept multiple thread accessing package info.

hotchemi:
When it comes to package info, yes. But I don't want to getTargetSdkVersion to return -1 unexpectedly.
I'm not sure volatile is enough...

shiraji:
OK. we can survive for the code. Let see if someone reports issue for `getTargetSdkVersion`.

hotchemi:
@shiraji Sorry, I changed my mind and delete it...! I guess it's hardly called at the same time:D

hotchemi:
#97 .

PR #102, opened 2016-02-20
Make source compatibility to 1.6.

hotchemi:
## Change

Kotlin 1.0 assume that they support Java 1.6 byte code until Android uses it.

refs. [Kotlin 1.0 Released: Pragmatic Language for JVM and Android | Kotlin Blog](https://blog.jetbrains.com/kotlin/2016/02/kotlin-1-0-released-pragmatic-language-for-jvm-and-android/)


aurae:
We've talked about this during DroidKaigi, I'm all for it, because it mitigates issues related to users on older JDK versions trying to include the library. However, we have to make sure that each release build is being created on JDK6. Could we use Java system properties verification in a unit test to ensure that we continue building on JDK6?

hotchemi:
Actually we can do that with travis ci option, but now we only test on the Java 1.8 to reduce the time of ci check. And I think, we don't have to check the interoperability for older JDK carefully because it's the role of Kotlin.

hotchemi:
But we should write something about JDK env to README actually.

shiraji:
Nice! This is the beauty of Kotlin. I wish I could discuss this with you guys at the event.

hotchemi:
Added an attention about jdk env.

Issue #59, opened 2015-12-10
Add embedded lint checks to prevent direct access to permission-protected methods


aurae:
This enhancement is derived from #56, in which the OP asked for a more convenient way to use PermissionDispatcher's delegate pattern. In my opinion, it would make sense to embed custom lint checks that prevent users from directly invoking methods annotated with `@NeedsPermission`. Furthermore, another check would raise a warning (preferably an error) if the call to the delegate object's `onRequestPermissionsResult()` method is omitted in the respective Activity / Fragment method.

```java
@NeedsPermission(Manifest.permission.CAMERA)
void useCamera() {
}

// Somewhere else...
void onCameraButtonClicked() {
    this.useCamera();    // This line would raise a lint error ("Trying to access permission-protected method directly", for instance)
}
```

[Futher reading (tools.android.com)](http://tools.android.com/tips/lint-custom-rules)

hotchemi:
Seems nice!

shiraji:
Are you guys expect lInt check done by Kotlin?

aurae:
Personally, I'm indifferent to the language used to implement the lint checks, but I'd welcome a contribution written in Kotlin, for sure!

shiraji:
One issue of writing lint in Kotlin is I don't know how to include kotlin
runtime.jar in lint.jar.

2016年1月15日(金) 20:22 Marcel Schnelle <notifications@github.com>:

aurae:
That's true, I haven't looked into that as well, which is why I was interested in that. Java works just as well for this task, though, seeing as all references keep to using Java for custom lint rules. So in case you are interested in implementing this, Java works just fine :+1:

shiraji:
OK, I will do this then in Java. I don't think this lint will be huge. We will be able to rewrite this in Kotlin if we find a way to include kotlin runtime in lint.jar.

aurae:
True. Thank you for the offer!

hotchemi:
Btw, can it be bundled with artifact?

shiraji:
Not sure how, yet but according to [this documentation](http://tools.android.com/tech-docs/new-build-system/aar-format), aar is able to include lint.jar.

hotchemi:
ic.

shiraji:
Maybe, we could use [the idea from Timber](https://github.com/JakeWharton/timber/blob/master/timber/build.gradle#L45-L55).

aurae:
[...]

PR #104, opened 2016-02-22
Add a lint for NeedsPermission and OnShowRationale

shiraji:
see #59 and its [comment](https://github.com/hotchemi/PermissionsDispatcher/issues/59#issuecomment-181818458)

### DIFF
--- a/lint/src/main/java/permissions/dispatcher/NoCorrespondingNeedsPermissionDetector.java
+++ b/lint/src/main/java/permissions/dispatcher/NoCorrespondingNeedsPermissionDetector.java
@@ -29,6 +29,16 @@ public class NoCorrespondingNeedsPermissionDetector extends Detector implements 
             Severity.ERROR,
             new Implementation(NoCorrespondingNeedsPermissionDetector.class, EnumSet.of(Scope.JAVA_FILE)));
 
+    static final Set<String> NEEDS_PERMISSION_NAME = new HashSet<String>() {{
+        add("NeedsPermission");
+        add("permissions.dispatcher.NeedsPermission");
+    }};
+
+    static final Set<String> ON_SHOW_RATIONALE_NAME = new HashSet<String>() {{
+        add("OnShowRationale");
+        add("permissions.dispatcher.OnShowRationale");
+    }};
+
     @Override
     public AstVisitor createJavaVisitor(JavaContext context) {
         return new AnnotationChecker(context);
@@ -36,22 +46,11 @@ public class NoCorrespondingNeedsPermissionDetector extends Detector implements 
 
     static class AnnotationChecker extends ForwardingAstVisitor {
         private final JavaContext context;
-        private Set<String> needsPermissionName;
-        private Set<String> onShowRationaleName;
-
         private Set<Annotation> needsPermissionAnnotations;
         private Set<Annotation> onShowRationaleAnnotations;
 
         private AnnotationChecker(JavaContext context) {
             this.context = context;
-
-            needsPermissionName = new HashSet<>();
-            needsPermissionName.add("NeedsPermission");
-            needsPermissionName.add("permissions.dispatcher.NeedsPermission");
-
-            onShowRationaleName = new HashSet<>();
-            onShowRationaleName.add("OnShowRationale");
-            onShowRationaleName.add("permissions.dispatcher.OnShowRationale");
 
             needsPermissionAnnotations = new HashSet<>();
             onShowRationaleAnnotations = new HashSet<>();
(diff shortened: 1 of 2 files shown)


lint/src/main/java/permissions/dispatcher/NoMatchingNeedsPermissionDetector.java
```diff
@@ -0,0 +1,113 @@
...
+import lombok.ast.AstVisitor;
+import lombok.ast.CompilationUnit;
+import lombok.ast.ForwardingAstVisitor;
+import lombok.ast.Node;
+
+public class NoMatchingNeedsPermissionDetector extends Detector implements Detector.JavaScanner {
+
+    public static final Issue ISSUE = Issue.create("NoMatchingNeedsPermission",
+            "No matching @NeedsPermission found",
+            "The processor would ignore @OnShowRationale annotation if there is no matching @NeedsPermissions. For instance, if there is @OnShowRationale(Manifest.permission.CAMERA), then there must be @NeedsPermission(Manifest.permission.CAMERA) in the same class.",
```
shiraji: For this short and long description is not good enough for this lint. I'm not sure how to express this lint...

lint/src/main/java/permissions/dispatcher/NoMatchingNeedsPermissionDetector.java
```diff
@@ -0,0 +1,113 @@
...
+            Category.CORRECTNESS,
+            4,
+            Severity.ERROR,
```
shiraji: To be honest, I don't know how to determine this lint is error or warning...

hotchemi:
Nice!

lint/src/main/java/permissions/dispatcher/NoMatchingNeedsPermissionDetector.java
shiraji: `s/matching/corresponding/` makes sense to me.

hotchemi:
Oh, sorry I didn't check the design but in this case, it causes compile error right?

lint/src/main/java/permissions/dispatcher/NoCorrespondingNeedsPermissionDetector.java
```diff
@@ -0,0 +1,113 @@
...
+            new Implementation(NoCorrespondingNeedsPermissionDetector.class, EnumSet.of(Scope.JAVA_FILE)));
+
+    @Override
+    public AstVisitor createJavaVisitor(JavaContext context) {
+        return new AnnotationChecker(context);
+    }
+
+    static class AnnotationChecker extends ForwardingAstVisitor {
+        private final JavaContext context;
+        private Set<String> needsPermissionName;
```
hotchemi: I guess `needsPermissionName` and `onShowRationaleName` can be initialized like below.

```java
private static final Set<String> NEEDS_PERMISSION_NAME = new HashSet<>() {{;
  add("OnShowRationale");
  add("permissions.dispatcher.OnShowRationale");
}};


```

lint/src/main/java/permissions/dispatcher/NoCorrespondingNeedsPermissionDetector.java
```diff
@@ -0,0 +1,113 @@
...
+
+            super.afterVisitCompilationUnit(node);
+        }
+
+        private boolean hasSameNodes(List<Node> first, List<Node> second) {
+            if (first.size() != second.size()) {
+                return false;
+            }
+
+            for (int i = 0; i < first.size(); i++) {
```
hotchemi: Below way is good for performance I guess.
```java
for (int i = 0, size = first.size(); i < size; i++) {
```

hotchemi:
Thx! one of the my concern is https://github.com/hotchemi/PermissionsDispatcher/pull/104#issuecomment-189031201 but other than that it's almost lgtm!

One question, How could we test it?

shiraji:
> Oh, sorry I didn't check the design but in this case, it causes compile error right?

No. PermissionsDispatcher does not generate onShowRationale method call if the `@OnShowRatinale` does not have corresponding `@NeedsPermission`

hotchemi:
Ah ic.

shiraji:
Let me squash commits after your review.

shiraji:
> One question, How could we test it?

I'm sorry. I accidentally ignore this one.

In terms of testing, since `LintDetectorTest` is not good enough for testing, yet, I don't think we can write unit testing for Lint at this moment. However, we can provide sample module that shows all lint warnings.

aurae:
`message`
*Useless @OnShowRationale declaration*

`summary`
*The method annotated with @OnShowRationale has no corresponding @NeedsPermission method, and will therefore be ignored by PermissionsDispatcher*

`explanation`
*The @OnShowRationale method with a certain signature is internally connected to another method annotated with @NeedsPermission and the same annotation value. Please ensure that there is a @NeedsPermission method with matching annotation values for this method.*

shiraji:
Nice! I will change wording.

hotchemi:
LInt master and English master are there:D

hotchemi:
Thx!

Issue #107, opened 2016-03-07
java.lang.RuntimeException: Unknown exception code: 1 msg null


kassim:
Getting the following crash reported in an app I'm working on
```
Caused by java.lang.RuntimeException: Unknown exception code: 1 msg null
       at android.os.Parcel.readException(Parcel.java:1484)
       at android.os.Parcel.readException(Parcel.java:1426)
       at android.app.ActivityManagerProxy.checkPermission(ActivityManagerNative.java:3450)
       at android.app.ContextImpl.checkPermission(ContextImpl.java:1703)
       at android.content.ContextWrapper.checkPermission(ContextWrapper.java:570)
       at android.content.ContextWrapper.checkPermission(ContextWrapper.java:570)
       at android.support.v4.content.PermissionChecker.checkPermission(PermissionChecker.java:89)
       at android.support.v4.content.PermissionChecker.checkSelfPermission(PermissionChecker.java:125)
       at permissions.dispatcher.PermissionUtils.hasSelfPermissions(PermissionUtils.java:40)
```

I'm not sure what the cause is but looking at this issue reported in the Appboy SDK https://github.com/Appboy/appboy-android-sdk/issues/42 their solution was to wrap the call in a try/catch

Do you know what may be the cause?

shiraji:
Could you give us a minimum project that reproduces this issue?

aurae:
Are you experiencing this consistently on a particular device? What's the OS version on it? This issue is most definitely related to the support library's `PermissionChecker` class. We'll have to check the Android source in order to find the root of this. As you said, both Appboy and PayPal wrap their `checkSelfPermission` invocation in a try/catch and that's it, which seems to be fine, however I'd like to find out what's causing this first.

kassim:
I can't reproduce it locally unfortunately. 
Lenovo A816 on 4.4.4 is the main device we're seeing it reported on, when retrieving Contacts permissions in my case.

aurae:
This is the offending implementation of `Parcel#readException`:

```java
/**
 * Use this function for customized exception handling.
 * customized method call this method for all unknown case
 * @param code exception code
 * @param msg exception message
 */
public final void readException(int code, String msg) {
    switch (code) {
        case EX_SECURITY:
            throw new SecurityException(msg);
        case EX_BAD_PARCELABLE:
            throw new BadParcelableException(msg);
        case EX_ILLEGAL_ARGUMENT:
[...]

PR #108, opened 2016-03-08
Add a workaround for RuntimeException when checking self permissions

shiraji:
see #107

### DIFF
--- a/library/src/main/android/permissions/dispatcher/PermissionUtils.java
+++ b/library/src/main/android/permissions/dispatcher/PermissionUtils.java
@@ -70,11 +70,30 @@ public final class PermissionUtils {
      */
     public static boolean hasSelfPermissions(Context context, String... permissions) {
         for (String permission : permissions) {
-            if (permissionExists(permission) && checkSelfPermission(context, permission) != PackageManager.PERMISSION_GRANTED) {
+            if (permissionExists(permission) && !hasSelfPermission(context, permission)) {
                 return false;
             }
         }
         return true;
+    }
+
+    /**
+     * Determine context has access to the given permission.
+     *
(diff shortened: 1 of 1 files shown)


library/src/main/android/permissions/dispatcher/PermissionUtils.java
```diff
@@ -78,6 +78,25 @@ public static boolean hasSelfPermissions(Context context, String... permissions)
...
+     *
+     * @param context context
+     * @param permission permission
+     * @return returns true if context has access to the given permission, false otherwise.
+     * @see #hasSelfPermissions(Context, String...)
+     */
+    private static boolean hasSelfPermission(Context context, String permission) {
+        try {
+            return checkSelfPermission(context, permission) == PackageManager.PERMISSION_GRANTED;
+        } catch (Throwable t) {
```
hotchemi: How about restricting only RuntimeException?

hotchemi:
Thx!

library/src/main/android/permissions/dispatcher/PermissionUtils.java
shiraji: Sure. Let me fix that. I was thinking just catch everything since this shouldn't happen.

hotchemi: Umm, ic. Not sure which is better but going on!

shiraji:
Fixed and squashed.

hotchemi:
:100:

library/src/main/android/permissions/dispatcher/PermissionUtils.java
shiraji: Yea!

shiraji:
Thanks!

Issue #133, opened 2016-04-11
Missing createAppSettingsIntent on 2.1.2


sibeliusseraphini:

sibeliusseraphini:
```
public static @Nullable Intent createAppSettingsIntent(Context context) {
        Intent intent;
        String packageName = context.getPackageName();
        if (Build.VERSION.SDK_INT >= GINGERBREAD) {
            // Utilize the dedicated Settings Action on API 9+
            intent = new Intent(Settings.ACTION_APPLICATION_DETAILS_SETTINGS);
            Uri uri = Uri.fromParts("package", packageName, null);
            intent.setData(uri);

        } else {
            // Compatibility implementation using a View intent and the native Settings app
            intent = new Intent(Intent.ACTION_VIEW);
            intent.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
            intent.setClassName("com.android.settings", "com.android.settings.InstalledAppDetails");
            String extraKey = Build.VERSION.SDK_INT == FROYO ? "pkg" : "com.android.settings.ApplicationPkgName";
            intent.putExtra(extraKey, packageName);
        }

        // If the intent can be used to resolve an Activity, return that intent. Otherwise, return null
        return context.getPackageManager().resolveActivity(intent, 0) != null ? intent : null;
    }
```

shiraji:
The method was deleted v2.0.3. Check this commit https://github.com/hotchemi/PermissionsDispatcher/commit/bb2537c91dac8287dbc24839033d8986980792b4

sibeliusseraphini:
@shiraji can we added it back? I think it is a very common use case

shiraji:
Could you give me a sample use case for this method? I don't see the point of having this method in this library at this moment. 

Yes, you are right. It is very common for Android app but not for library.

PR #134, opened 2016-04-11
fix hotchemi/PermissionsDispatcher#133, adding createAppSettingsInten…

sibeliusseraphini:
…t method

### DIFF
--- a/library/src/main/android/permissions/dispatcher/PermissionUtils.java
+++ b/library/src/main/android/permissions/dispatcher/PermissionUtils.java
@@ -3,12 +3,18 @@ package permissions.dispatcher;
 import android.annotation.TargetApi;
 import android.app.Activity;
 import android.content.Context;
+import android.content.Intent;
 import android.content.pm.PackageInfo;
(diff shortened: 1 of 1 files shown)


shiraji:
Could you give us a sample use case for this method? I don't see how this method is useful for this library. (Yes, it is useful for Android app, but not for library)

sibeliusseraphini:
```java
@OnNeverAskAgain(Manifest.permission.CAMERA)
public void neverAskAgain() {
        if (getView() != null) {
            final Snackbar snackBar = Snackbar.make(getView(), R.string.msg_permission_denied, Snackbar.LENGTH_LONG);
            snackBar.setAction("Habilitar", new View.OnClickListener() {
                public void onClick(View v) {

                    Intent intent = PermissionUtils.createAppSettingsIntent(getActivity());
                    PermissionFragment.this.startActivity(intent);
                }
            });
            snackBar.show();
        }
    }
```

sibeliusseraphini:
this method will almos always be used together with @OnNeverAskAgain

PR #188, opened 2016-08-19
Endorse AA plugin in README

### LINKED ISSUE #168: Endorse AA plugin in README
Let's consider adding a section on `AndroidAnnotations` in the README, in which we redirect people to [Aleksander's plugin](https://github.com/AleksanderMielczarek/AndroidAnnotationsPermissionsDispatcherPlugin). This could be generalized into an "Integration" section in the future once we add support for more frameworks, but since AA is the only one right now, we should add it like that.

What do you think, @hotchemi @shiraji?

LouisCAD:
This commit also fixes a typo (missing s).

resolves #168


hotchemi:
Awesome 😄

Issue #192, opened 2016-08-25
maxSdkVersion support


hotchemi:
ref: https://developer.android.com/guide/topics/manifest/uses-permission-element.html

### Overview

`<uses-permission>` element has an attribute which is called maxSdkVersion.
For example `getExternalFilesDir(String)` and `getExternalCacheDir()` require READ_EXTERNAL_STORAGE permission when api lever is 19 or lower.
So in that case it's better to not to ask permission granting. Just deal with "granted".

PR #204, opened 2016-09-24
Support maxSdkVersion

hotchemi:
ref: https://github.com/hotchemi/PermissionsDispatcher/issues/192
ref: https://developer.android.com/guide/topics/manifest/uses-permission-element.html

## Change

- Add maxSdkVersionSupport.
  - Add `maxSdkVersion` in NeedsPermission.java
  - If specified value is above 0, processor creates code that only checks SDK_INT and call needs permission method.
  - If specified value is 0 or below, it does nothing.


hotchemi:
@shiraji @aurae Can you take a look and merge?

hotchemi:
Let me merge!

PR #231, opened 2016-12-09
Fix the problem with lint check with new line

### LINKED ISSUE #230: NeedOnRequestPermissionsResult lint check is firing erroneously
After https://github.com/hotchemi/PermissionsDispatcher/pull/227 was merged and I bumped the version in app, it looks like there is a misfiring lint check.

This is the check — NeedOnRequestPermissionsResult: Call the "onRequestPermissionsResult" method of the generated PermissionsDispatcher class in the respective method of your Activity or Fragment

And here is code that it is triggering on:
```java
    @Override
    public void onRequestPermissionsResult(int requestCode, String[] permissions,
            int[] grantResults) {
        super.onRequestPermissionsResult(requestCode, permissions, grantResults);
        FooBarFragmentPermissionsDispatcher
                .onRequestPermissionsResult(this, requestCode, grantResults);
    }
```

So the call to the PermissionsDispatcher is actually happening but the lint check does not think it is.

cc @eyedol @hotchemi @felipecsl

shiraji:
Close #230 

Since I'm not familiar with new api, I'm not sure this is how the lint check method call...but it should work in any code format.


hotchemi:
👀

shiraji:
I have no idea how to test lint with generated code...

Do you have any idea @eyedol ?

lint/src/main/java/permissions/dispatcher/CallOnRequestPermissionsResultDetector.java
```diff
@@ -110,14 +111,37 @@ public void visitMethod(PsiMethod method) {
     }
 
     private static boolean checkMethodCall(PsiMethod method, PsiClass psiClass) {
-        // FIXME: I'm sure there is a better way of checking if the on onRequestPermissionsResult
-        // method from the generated class is being called.
         PsiCodeBlock codeBlock = method.getBody();
+        if (codeBlock == null) {
+            return false;
```
eyedol: Nice one

hotchemi:
LGTM 💯

Issue #211, opened 2016-09-27
Mechanism that delegate the check work to users


hotchemi:
ref: https://github.com/hotchemi/PermissionsDispatcher/issues/187

This is definitely unexpected, but some devices like from Xiaomi need different check flow to get the state of permission.
Currently we provide `PermisisonUtils` in library module, but I'm thinking to provide the following interface to allow users to put in their original logic.

``` java
@RuntimePermission(checker = XiaomiChecker.class) // default value is PermisisonUtils.class
class Some {
}
```

@aurae @shiraji What do you think?

@a1018875550 Can you solve the problem with above interface?

a1018875550:
It's a good idea,but it will change the processor code;I will try to achieve it for Java

hotchemi:
Umm? in that case all you have to do is to create your own checher in Java and just inject it.

a1018875550:
Do u just need the `XiaomiChecker.class`?

hotchemi:
If we will impl that yes, user's task is to create checker class on app side and just inject it.

aurae:
I'm not against the idea of generalizing the "PermissionChecker" implementation per se, however I'm curious as to what the generalized interface would look like? @a1018875550, can you explain in greater detail the differences of Xiaomi's implementation for runtime permissions, compared to stock Android? Specifically, how do rationales and the "never ask again" mechanism work?

I'm picturing an extension library that provides Xiaomi compatibility, which may be added as an additional dependency on top of PermissionsDispatcher. When done right, this should not interfere with the existing processor code in place for Java, however my confusion regarding how Xiaomi deals with this remains.

shiraji:
@aurae are you saying we are going to generate `PermissionUtils` class at compile time? How to accomplish extension mechanism? It still depends on what @a1018875550 answer but I'm interested to your idea.

hotchemi:
I guess the only difference is here. We have to use `AppOpsManager` to get the state of permission correctly, right? @a1018875550

https://github.com/a1018875550/PermissionDispatcher/blob/master/library/src/main/android/org.jokar.permissiondispatcher.library/PermissionUtils.java#L128

hotchemi:
And one more reason I proposed the feature that delegates the check to user is we can deal with similar problem like particular device from **awesome** hardware manufacturer would have a different permission mechanism...
I hope it won't occur but this is Android platform, it's probable...so in that case we don't want to distribute each special library right? The simplest solution is delegation itself.

[...]

PR #235, opened 2016-12-13
Add workaround for Xiaomi devices.

hotchemi:
ref: https://github.com/hotchemi/PermissionsDispatcher/issues/211

- This change will add workaround for Xiaomi.
  - I referenced this method actually. https://github.com/a1018875550/PermissionDispatcher/blob/master/library/src/main/android/org.jokar.permissiondispatcher.library/PermissionUtils.java#L128

### DIFF
--- a/library/src/main/android/permissions/dispatcher/PermissionUtils.java
+++ b/library/src/main/android/permissions/dispatcher/PermissionUtils.java
@@ -110,7 +110,7 @@ public final class PermissionUtils {
     private static boolean hasSelfPermissionForXiaomi(Context context, String permission) {
         AppOpsManager appOpsManager = (AppOpsManager) context.getSystemService(Context.APP_OPS_SERVICE);
         int checkOp = appOpsManager.checkOp(AppOpsManager.permissionToOp(permission), Process.myUid(), context.getPackageName());
-        return checkOp == AppOpsManager.MODE_ALLOWED;
+        return checkOp == AppOpsManager.MODE_ALLOWED && ActivityCompat.checkSelfPermission(context, permission) == PackageManager.PERMISSION_GRANTED;
     }
 
     /**
--- a/library/src/main/android/permissions/dispatcher/PermissionUtils.java
+++ b/library/src/main/android/permissions/dispatcher/PermissionUtils.java
@@ -2,10 +2,12 @@ package permissions.dispatcher;
 
 import android.annotation.TargetApi;
 import android.app.Activity;
+import android.app.AppOpsManager;
 import android.content.Context;
 import android.content.pm.PackageInfo;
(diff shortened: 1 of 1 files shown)


hotchemi:
@a1018875550 @shiraji @aurae Please review!

@a1018875550 Do we need to check both `ActivityCompat.checkSelfPermission` and `AppOpsManager`?

library/src/main/android/permissions/dispatcher/PermissionUtils.java
```diff
@@ -94,13 +96,23 @@ public static boolean hasSelfPermissions(Context context, String... permissions)
      * @see #hasSelfPermissions(Context, String...)
      */
     private static boolean hasSelfPermission(Context context, String permission) {
+        if (Build.VERSION.SDK_INT > Build.VERSION_CODES.M && Build.MANUFACTURER.equalsIgnoreCase("Xiaomi")) {
```
aurae: Is there any other way to detect a Xiaomi device than to use `Build.MANUFACTURER` here? I just want to make sure we can catch all of them, and that there aren't any weird devices with an identifier like `xia` or something.

a1018875550:
yes ,need  check both ActivityCompat.checkSelfPermission and AppOpsManager on "Xiaomi" phone;

library/src/main/android/permissions/dispatcher/PermissionUtils.java
hotchemi: Umm, I actually I don't have Xiaomi device so can't test it...just referenced https://github.com/hotchemi/PermissionsDispatcher/issues/211#issuecomment-265660654.
@a1018875550 Any idea?

hotchemi:
Addressed.

library/src/main/android/permissions/dispatcher/PermissionUtils.java
a1018875550: sorry，I only know this way to distinguish the device

hotchemi:
Okay to merge?🙂

library/src/main/android/permissions/dispatcher/PermissionUtils.java
aurae: The internet seems to agree that this is sufficient. I guess there will be issues opened in case there actually _is_ such a device. I'm fine with this for now!

aurae:
LGTM!

library/src/main/android/permissions/dispatcher/PermissionUtils.java
hotchemi: @aurae Thx for your dedicated investigation:D

hotchemi:
Finally!

PR #242, opened 2016-12-18
Handle the case Build.MANUFACTURER = null

shiraji:
I hope it doesn't happen in real world, but if `Build.MANUFACTURER` is null, the condition give us NullPointerException. We should avoid the unexpected Exception.

### DIFF
--- a/library/src/main/android/permissions/dispatcher/PermissionUtils.java
+++ b/library/src/main/android/permissions/dispatcher/PermissionUtils.java
@@ -96,7 +96,7 @@ public final class PermissionUtils {
      * @see #hasSelfPermissions(Context, String...)
      */
     private static boolean hasSelfPermission(Context context, String permission) {
-        if (Build.VERSION.SDK_INT > Build.VERSION_CODES.M && Build.MANUFACTURER.equalsIgnoreCase("Xiaomi")) {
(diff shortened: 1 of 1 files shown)


hotchemi:
LGTM 💯

library/src/main/android/permissions/dispatcher/PermissionUtils.java
```diff
@@ -96,7 +96,7 @@ public static boolean hasSelfPermissions(Context context, String... permissions)
      * @see #hasSelfPermissions(Context, String...)
      */
     private static boolean hasSelfPermission(Context context, String permission) {
-        if (Build.VERSION.SDK_INT > Build.VERSION_CODES.M && Build.MANUFACTURER.equalsIgnoreCase("Xiaomi")) {
+        if (Build.VERSION.SDK_INT > Build.VERSION_CODES.M && "Xiaomi".equalsIgnoreCase(Build.MANUFACTURER)) {
```
a1018875550: sorry,the `Build.VERSION.SDK_INT > Build.VERSION_CODES.M` should be `Build.VERSION.SDK_INT >= Build.VERSION_CODES.M`;It's should be work in Build.VERSION_CODES.M and above

aurae:
LGTM as well, but I strongly believe this won't be an issue.

PR #239, opened 2016-12-17
Use JFrog's artifactory plugin to upload snaphots

eyedol:
Fixes #232 

Changes here include:

- Add a deploy script that uploads snapshots of master to jfrog
- Add SNAPSHOT to current version number of the library

Todo from project maintainer

- [ ] Create an OJO account. 
- [ ] Make a snapshot release.
- [x] Unbreak test.

Since you have a bintray account already you don't have to manually create the OJO account. Follow [this](https://www.jfrog.com/confluence/display/RTF/Deploying+Snapshots+to+oss.jfrog.org#DeployingSnapshotstooss.jfrog.org-GettingStartedwithOJO) It's some clicks away.

Issue #27, opened 2015-09-07
Separate annotation and util module


hotchemi:
Like that.
https://github.com/MichaelEvans/Aftermath

aurae:
We already have a setup like that, kind of, since the library module contains two source folders for `android` and plain `java`. What would be the advantage of this?

hotchemi:
Nothing especially but when I tried to do that, it didn't go well may be...

hotchemi:
It's worth to re-challenge.

hotchemi:
This issue includes writing unit test for `PermissionUtis` with Robolectric and mockito.

hotchemi:
too old

PR #243, opened 2016-12-19
Add workaround for processor test phase

hotchemi:
### Overview

- This pr fixes https://github.com/hotchemi/PermissionsDispatcher/pull/239 build failure.
  - Currently we have a tricky build phase that copies library jar to processor lib folder to resolve dependency.
  - But only we need is just PermissionUtil class interface so that I create a dummy `PermissionUtils' class and remove copy task.
  - It's still a bit tricky but it becomes much more better...it'd be completely resolved on https://github.com/hotchemi/PermissionsDispatcher/issues/27 actually.


hotchemi:
@aurae I'm not sure but it seems test cases under `testV13` are not executed by `./gradlew check`?

shiraji:
> it seems test cases under testV13 are not executed by ./gradlew check?

I couldn't dig into deeply but it doesn't.

hotchemi:
Thx! Let me merge.

Issue #244, opened 2016-12-20
Device is XiaoMi ,Check the default permissions  java.lang.IllegalArgumentException: Unknown operation string: null


lftx:
```
AndroidRuntime: FATAL EXCEPTION: main
                                                                             Process: permissions.dispatcher.sample, PID: 8940
                                                                             java.lang.IllegalArgumentException: Unknown operation string: null
                                                                                 at android.app.AppOpsManager.strOpToOp(AppOpsManager.java:1464)
                                                                                 at android.app.AppOpsManager.checkOp(AppOpsManager.java:1486)
                                                                                 at permissions.dispatcher.PermissionUtils.hasSelfPermissionForXiaomi(PermissionUtils.java:115)
                                                                                 at permissions.dispatcher.PermissionUtils.hasSelfPermission(PermissionUtils.java:101)
                                                                                 at permissions.dispatcher.PermissionUtils.hasSelfPermissions(PermissionUtils.java:81)
                                                                                 at permissions.dispatcher.sample.MainActivityPermissionsDispatcher.showNetWorkWithCheck(MainActivityPermissionsDispatcher.java:52)
                                                                                 at permissions.dispatcher.sample.MainActivity.onClick(MainActivity.java:41)
                                                                                 at android.view.View.performClick(View.java:5207)
                                                                                 at android.view.View$PerformClick.run(View.java:21177)
                                                                                 at android.os.Handler.handleCallback(Handler.java:739)
                                                                                 at android.os.Handler.dispatchMessage(Handler.java:95)
                                                                                 at android.os.Looper.loop(Looper.java:148)
                                                                                 at android.app.ActivityThread.main(ActivityThread.java:5458)
                                                                                 at java.lang.reflect.Method.invoke(Native Method)

```

Modify ：

```java
 private static boolean hasSelfPermissionForXiaomi(Context context, String permission) {
        AppOpsManager appOpsManager = (AppOpsManager) context.getSystemService(Context.APP_OPS_SERVICE);
        String op = AppOpsManager.permissionToOp(permission);
        if (!TextUtils.isEmpty(op)) {
            int checkOp = appOpsManager.checkOp(op, Process.myUid(), context.getPackageName());
            return checkOp == AppOpsManager.MODE_ALLOWED && ActivityCompat.checkSelfPermission(context, permission) == PackageManager.PERMISSION_GRANTED;
        }
        return true;
    }
```

shiraji:
Here is the better format of above code.

[...]

PR #250, opened 2016-12-25
Fix to use compat class.

hotchemi:
## Issues

- closes https://github.com/hotchemi/PermissionsDispatcher/issues/244

## Change

- Just in case user select the normal permission, fix to return true when `permissionToOp` returns null
- And change to use `AppOpsManagerCompat` instead of real `AppOpsManager`

### DIFF
--- a/library/src/main/android/permissions/dispatcher/PermissionUtils.java
+++ b/library/src/main/android/permissions/dispatcher/PermissionUtils.java
@@ -1,14 +1,13 @@
 package permissions.dispatcher;
 
-import android.annotation.TargetApi;
 import android.app.Activity;
-import android.app.AppOpsManager;
 import android.content.Context;
 import android.content.pm.PackageInfo;
 import android.content.pm.PackageManager;
 import android.os.Build;
 import android.os.Process;
 import android.support.v4.app.ActivityCompat;
+import android.support.v4.app.AppOpsManagerCompat;
 import android.support.v4.util.SimpleArrayMap;
 
 import static android.support.v4.content.PermissionChecker.checkSelfPermission;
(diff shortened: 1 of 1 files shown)


library/src/main/android/permissions/dispatcher/PermissionUtils.java
```diff
@@ -135,7 +137,6 @@ public static boolean shouldShowRequestPermissionRationale(Activity activity, St
      * @param context context
      * @return target sdk version
      */
-    @TargetApi(Build.VERSION_CODES.DONUT)
```
hotchemi: I guess this is no longer needed.

hotchemi: I guess this is not needed.

library/src/main/android/permissions/dispatcher/PermissionUtils.java
```diff
@@ -106,11 +105,14 @@ private static boolean hasSelfPermission(Context context, String permission) {
...
     private static boolean hasSelfPermissionForXiaomi(Context context, String permission) {
-        AppOpsManager appOpsManager = (AppOpsManager) context.getSystemService(Context.APP_OPS_SERVICE);
-        int checkOp = appOpsManager.checkOp(AppOpsManager.permissionToOp(permission), Process.myUid(), context.getPackageName());
-        return checkOp == AppOpsManager.MODE_ALLOWED && checkSelfPermission(context, permission) == PackageManager.PERMISSION_GRANTED;
+        String permissionToOp = AppOpsManagerCompat.permissionToOp(permission);
+        if (permissionToOp == null) {
+            // in case of normal permissions(e.g. INTERNET)
+            return true;
+        }
+        int noteOp = AppOpsManagerCompat.noteOp(context, permissionToOp, Process.myUid(), context.getPackageName());
```
shiraji: I'm quite not sure but are we gonna ignore the case `noteOp` throws `SecurityException`. I know there is no `noteOpNoThrow` method.

hotchemi: Good catch, actually I'm not sure how `SecurityException` is thrown either...
I suppose it's okay since nobody has complained about it yet...YAGNI:D

shiraji: Yea, that's what I thought.

hotchemi:
Thx for your review!

PR #260, opened 2017-01-04
Document hidden features

shiraji:
Well, as you guys know already, I need English teachers to fix the "syntax".


hotchemi:
Thx! To make README more legible and approachable let me split some part to wiki later.
Let me merge!

PR #263, opened 2017-01-08
Change how to create the generated class package name

### LINKED ISSUE #262: PermissionsDispatcher makes wrong class hierarchy
When I set @RuntimePermission and related annotations in the nested class, PermissionDispatcher generates classes with wrong class hierarchy.    
If there's a code like this,
```
package test;
public class Test {
  //Not an Activity
  @RuntimePermission
  public static class NestedActivity extends AppCompatActivity {
    //An Activity
    @NeedsPermission("android.permission.ACCESS_FINE_LOCATION")//Some permissions for this method
    public void permissionRequiredMethod(){
       //Some code here
    }
  }
}
```
From this code, PermissionDispatcher generates a code like this:
```
package test.Test;
public class NestedActivityPermissionDispatcher{
  //Generated code here
}
```
This generated code makes some error where referencing Test class and package-private methods in generated code. (This error doesn't happen at my code.)    
    
I think generated code must be like this to work correctly:
```
package test;
public class Test$NestedActivityPermissionDispatcher{
  //Generated code here
}
```

shiraji:
Close #262

### DIFF
--- a/processor/src/main/kotlin/permissions/dispatcher/processor/util/Extensions.kt
+++ b/processor/src/main/kotlin/permissions/dispatcher/processor/util/Extensions.kt
@@ -7,15 +7,21 @@ import permissions.dispatcher.OnShowRationale
 import permissions.dispatcher.processor.TYPE_UTILS
 import javax.lang.model.element.Element
 import javax.lang.model.element.ExecutableElement
+import javax.lang.model.element.PackageElement
 import javax.lang.model.element.TypeElement
 import javax.lang.model.type.TypeMirror
 
 /**
  * Returns the package name of a TypeElement.
  */
-fun TypeElement.packageName(): String {
-    val qn = this.qualifiedName.toString()
-    return qn.substring(0, qn.lastIndexOf('.'))
+fun TypeElement.packageName() = enclosingElement.packageName()
+
+private fun Element?.packageName(): String {
+    return when (this) {
+        is TypeElement -> packageName()
+        is PackageElement -> qualifiedName.toString()
+        else -> this?.enclosingElement?.packageName() ?: ""
+    }
 }
 
 /**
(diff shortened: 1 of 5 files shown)


processor/src/main/kotlin/permissions/dispatcher/processor/util/Extensions.kt
```diff
@@ -7,15 +7,20 @@ import permissions.dispatcher.OnShowRationale
...
  * Returns the package name of a TypeElement.
  */
 fun TypeElement.packageName(): String {
-    val qn = this.qualifiedName.toString()
-    return qn.substring(0, qn.lastIndexOf('.'))
+    val enclosingElement = enclosingElement
+    return when (enclosingElement) {
+        is TypeElement -> enclosingElement.packageName()
+        is PackageElement -> enclosingElement.qualifiedName.toString()
+        else -> ""
```
shiraji: I did this, sInce [this line](https://github.com/hotchemi/PermissionsDispatcher/blob/master/processor/src/main/kotlin/permissions/dispatcher/processor/RuntimePermissionsElement.kt#L16) comes before `init` block, TypeElement hasn't validated, yet. If `enclosingElement` returns non-`TypeElement`/`PackageElement`, I believe some validation fails, later.

aurae: In these cases, would it be feasible to walk up the "enclosingElement" hierarchy to check for `TypeElement`/ `PackageElement` higher up the chain?

shiraji: Make sense. I was expecting `null` here. Let me fix this.

hotchemi:
@shiraji Thx! Before review can we wait a conclusion of my question? https://github.com/hotchemi/PermissionsDispatcher/issues/262#issuecomment-271240627
As I wrote, I'm not for allowing user to do a wrong approach so let me clarify the pros of the change.

shiraji:
i will add and rebase the commit as soon as possible.

shiraji:
done.

hotchemi:
Thx!!!

aurae:
Great job!

PR #284, opened 2017-03-05
Add compile-time check for NeedsPermission's overloaded methods

### LINKED ISSUE #283: Add compile-time check for overloaded methods
As discussed in #282, rather than having to modify our generated method names in the events of an overloaded method, we should enforce an additional compile-time check that fails the build if a user chooses to use it with their `@NeedsPermission` methods:

```java
@RuntimePermissions
class SomeFragment extends Fragment {
    @NeedsPermission(Manifest.permission.ACCESS_FINE_LOCATION)
    void showMap(String json) {
        // ...
    }

    @NeedsPermission(Manifest.permission.ACCESS_FINE_LOCATION)
    void showMap() {
        // ...
    }
}
```

The precondition for this check would be quite simple: If the `RuntimePermissionsElement` finds duplicates in the list of annotated `@NeedsPermission` methods, we can assume that method overloading is taking place. We should raise an Exception that tells the user what's wrong with that - maybe something like `AmbiguousMethodException`?

shiraji:
Close #283

### DIFF
--- a/processor/src/main/kotlin/permissions/dispatcher/processor/RuntimePermissionsElement.kt
+++ b/processor/src/main/kotlin/permissions/dispatcher/processor/RuntimePermissionsElement.kt
@@ -35,6 +35,7 @@ class RuntimePermissionsElement(e: TypeElement) {
         checkPrivateMethods(needsElements, NeedsPermission::class.java)
         checkMethodSignature(needsElements)
         checkMixPermissionType(needsElements, NeedsPermission::class.java)
+        checkDuplicatedMethodName(needsElements)
     }
 
     private fun validateRationaleMethods() {
--- a/processor/src/main/kotlin/permissions/dispatcher/processor/exception/DuplicatedMethodNameException.kt
+++ b/processor/src/main/kotlin/permissions/dispatcher/processor/exception/DuplicatedMethodNameException.kt
@@ -1,0 +1,6 @@
+package permissions.dispatcher.processor.exception
+
+import permissions.dispatcher.processor.util.simpleString
+import javax.lang.model.element.ExecutableElement
+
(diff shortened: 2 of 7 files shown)


hotchemi:
👀

processor/src/main/kotlin/permissions/dispatcher/processor/exception/AmbiguousMethodException.kt
```diff
@@ -0,0 +1,6 @@
+package permissions.dispatcher.processor.exception
+
+import permissions.dispatcher.processor.util.simpleString
+import javax.lang.model.element.ExecutableElement
+
+class AmbiguousMethodException(e: ExecutableElement) : RuntimeException("'${e.simpleString()}()' has duplicated '@NeedsPermission' method. The method annotated with '@NeedsPermission' must has the unique name.")
```
hotchemi: How about `DuplicatedMethodNameException`? From my perspective it's a bit intuitive.

processor/src/main/kotlin/permissions/dispatcher/processor/RuntimePermissionsElement.kt
```diff
@@ -35,6 +35,7 @@ class RuntimePermissionsElement(e: TypeElement) {
         checkPrivateMethods(needsElements, NeedsPermission::class.java)
         checkMethodSignature(needsElements)
         checkMixPermissionType(needsElements, NeedsPermission::class.java)
+        checkMethodOverloads(needsElements)
```
hotchemi: How about `checkDuplicatedMethodName`?

hotchemi:
@shiraji Sorry for the late, can you check? 🙇

shiraji:
Fixed and squashed! (during session break!)

hotchemi:
LGTM 💯

aurae:
Thanks!

Issue #299, opened 2017-04-13
Callback invoked on old instance of fragment after rotation while "Allow app to {user permission}" is shown.


iaypym:
Hi there!

I have following issue with permission dispatcher:
**Current behavior**
Permission dispatcher invokes callback on old instance of fragment. 
It happens with following steps:
1. FragmentNameFragmentPermissionsDispatcher.callbackMethodWithCheck is called
2. System dialog (something like "Allow this app to read contacts") is shown to user.
3. User rotate device.
4. User clicks on Allow
5. callbackMethod is called on old instance of fragment

**Expected behavior:**
Permission dispatcher invokes callback on current fragment instance.

**Requirements**
It happens only when callbackMethod have following declaration:
```java
@NeedsPermission(READ_CONTACTS)
void callbackMethod(Bundle bundle);
```
i.e. callbackMethod must have bundle argument.
This behavior observed on LG K10 (Android 7.0), Huawei P8 Lite(Android 7.0), Genymotion emulator (Android 7.0)
PermissionDispatcher version - 2.1.3 and 2.3.2

PermissionDispatcher generates following code for onRequestPermissionsResult:
```java
static void onRequestPermissionsResult(FragmentName target, int requestCode, int[] grantResults) {
    switch (requestCode) {
      case REQUEST_CALLBACKMETHOD:
      if (PermissionUtils.getTargetSdkVersion(target.getActivity()) < 23 && !PermissionUtils.hasSelfPermissions(target.getActivity(), PERMISSION_CALLBACKMETHOD)) {
        target.showPermissionDenied();
        return;
      }
      if (PermissionUtils.verifyPermissions(grantResults)) {
        if (PENDING_CALLBACKMETHOD != null) {
          PENDING_CALLBACKMETHOD.grant();
        }
      } else {
        if (!PermissionUtilsV13.getInstance().shouldShowRequestPermissionRationale(target, PERMISSION_CALLBACKMETHOD)) {
          target.showNeverAskAgainReadContactsPermission();
        } else {
          target.showPermissionDenied();
        }
      }
      PENDING_CALLBACKMETHOD = null;
      break;
      default:
      break;
    }
  }
```

hotchemi:
@iaypym Thx for your report. Can I ask the reason of below?

> i.e. callbackMethod must have bundle argument.

iaypym:
[...]

Issue #304, opened 2017-04-28
Make android.app.Fragment handling way more explicitly


hotchemi:
## Overview

Now we have the problem(https://github.com/hotchemi/PermissionsDispatcher/issues/279) that is caused by `optional-base` plugin and new Gradle. And I also found out that there's an unresolvable case like below with current approach:

- When user want to use both `v4.Fragment` and normal `Fragment` in the same app with v13 support library dependency

To avoid that I'd like to propose the following interface change.

- Remove optional-base gradle plugin
- Add v13 argument in `@RuntimePermission`, if its value is true the library generates a code with `FragmentCompat`.

```java
@Target(ElementType.TYPE)
@Retention(RetentionPolicy.CLASS)
public @interface RuntimePermissions {
  boolean v13() default false;
}
```

I'd say that by changing the way of dealing with Fragment explicitly we'll be able to avoid unexpected bugs eternally and make code much more declarative.

What do you think? @aurae @shiraji

aurae:
I'm unsure why using both `Fragment` variants together would become an issue. Could you elaborate on that? As far as I remember, the different `ProcessorUnit` implementations would apply to only their respective variant, so we can't get into a state where they get mixed up.

I agree that we should get rid of the Optional plugin - at this point, there are cleaner ways to achieve optional dependencies. I'm still under the impression that we can make the Class lookup work dynamically, without having to require our users to modify their `RuntimePermissions` annotations. Ideally, the processor would do that automatically, but alas...

For now, *I'm fine with adding the `v13` argument*, as we do have a significant amount of users that rely on this feature, and we should mitigate the problem in a short-term way. However, I'm taking the liberty to experiment a bit more with the new possibilities of Gradle, trying to get a deeper understanding of how its classpath might've changed in a recent version.

shiraji:
I have some questions. Since the plugin sit here before I contribute to this project, I'm not sure the issue here.

* Are you sure [optional-base](https://github.com/nebula-plugins/gradle-extra-configurations-plugin) cause the error for https://github.com/hotchemi/PermissionsDispatcher/issues/279#issuecomment-295184659 ?
* It seems we use older version of the plugin. Can we upgrade this plugin?
* Isn't it possible to fix the plugin? I can contribute it if you want.

Personally, I don't like adding `v13` argument because the developer requires to think additional step to handling wether the Fragment is v13 or not. And the most developer will tend to forget about this feature. It leads to raise issues 🙅‍♂️

[...]

Issue #305, opened 2017-04-29
Remove a workaround for targetSdkVersion < 23


hotchemi:
## Overview

- Currently we have a workaround that checks `PermissionUtils.getTargetSdkVersion(target) < 23` to make sure that the library works under Marshmallow device.
- This feature is requested by a few user initially, but it's over one and a half year since the library came out so it's okay to remove this workaround to avoid unnecessary check for most of user.

```java
    static void onRequestPermissionsResult(MainActivity target, int requestCode, int[] grantResults) {
        switch (requestCode) {
            case REQUEST_SHOWCAMERA:
                if (PermissionUtils.getTargetSdkVersion(target) < 23 && !PermissionUtils.hasSelfPermissions(target, PERMISSION_SHOWCAMERA)) {
                    target.onCameraDenied();
                    return;
                }
                if (PermissionUtils.verifyPermissions(grantResults)) {
                    target.showCamera();
                } else {
                    if (!PermissionUtils.shouldShowRequestPermissionRationale(target, PERMISSION_SHOWCAMERA)) {
                        target.onCameraNeverAskAgain();
                    } else {
                        target.onCameraDenied();
                    }
                }
                break;
            default:
                break;
        }
    }
```

aurae:
I feel like, at this point, we can force our users to use a `targetSdkVersion` of at least 23. I'm interested whether we can check this in the annotation processor, actually. I'd like to raise an Exception on lower Target SDK levels, just because we can avoid users opening issues regarding a regression in support.

hotchemi:
Umm, I'd say it's a bit strong restriction. 
The reason of checking getTargetSdkVersion is to deal with the case that user turn off a permission from setting page on under Marshmallow device, but I think it's too rare case to take care.
In other words about 99% case PermissionsDispatcher assures the correct action even if taretSdkVersion is lower than 23. Thus showing a caution message or something is fine for sure 😺

aurae:
Ah, I see now. However, if we keep this conditional logic anyway (to show the "caution message" you're referring to, for instance), why not just keep the current functionality? I agree that we can could clean up the generated code here if we removed the `targetSdkVersion` check, but if we just remove the behavior and keep the conditional, I'd prefer the current way.

hotchemi:
[...]

Issue #279, opened 2017-02-21
permissions.dispatcher.processor.exception.SupportV13MissingException with gradle plugin  2.4.0-dev


ivangonzalez88:
## Overview

Hi,

I have an issue with permission dispatcher, it is throwing the below exception even if i add a dependency for support-v13. The issue happens when i use the following:

gradle plugin  version 
com.android.tools.build:gradle:2.4.0-dev 
wrapper
distributionUrl=https\://services.gradle.org/distributions/gradle-3.3-all.zip

Everything works fine on 'com.android.tools.build:gradle:2.2.3', but i actually depend on de 2.4.0-dev version.

Caused by: permissions.dispatcher.processor.exception.SupportV13MissingException: PermissionsDispatcher for annotated class 'AlertActivationFragment' can't be generated, because the support-v13 dependency is missing on your project
	at permissions.dispatcher.processor.impl.NativeFragmentProcessorUnit.checkPrerequisites(NativeFragmentProcessorUnit.kt:25)
	at permissions.dispatcher.processor.impl.BaseProcessorUnit.createJavaFile(BaseProcessorUnit.kt:34)
	at permissions.dispatcher.processor.PermissionsProcessor.process(PermissionsProcessor.kt:70)
	at com.sun.tools.javac.processing.JavacProcessingEnvironment.callProcessor(JavacProcessingEnvironment.java:794)
	at com.sun.tools.javac.processing.JavacProcessingEnvironment.discoverAndRunProcs(JavacProcessingEnvironment.java:705)
	at com.sun.tools.javac.processing.JavacProcessingEnvironment.access$1800(JavacProcessingEnvironment.java:91)
	at com.sun.tools.javac.processing.JavacProcessingEnvironment$Round.run(JavacProcessingEnvironment.java:1035)
	at com.sun.tools.javac.processing.JavacProcessingEnvironment.doProcessing(JavacProcessingEnvironment.java:1176)
	at com.sun.tools.javac.main.JavaCompiler.processAnnotations(JavaCompiler.java:1170)
	at com.sun.tools.javac.main.JavaCompiler.compile(JavaCompiler.java:856)
	at com.sun.tools.javac.main.Main.compile(Main.java:523)

### Expected

 Be able to compile and run.

### Actual

 not compiling, annotation processing not working

## Environment

gradle plugin  version 
com.android.tools.build:gradle:2.3.0-dev 
wrapper
[...]

PR #306, opened 2017-04-30
Remove optional-base plugin

hotchemi:
## Issues
- Resolves https://github.com/hotchemi/PermissionsDispatcher/issues/299, https://github.com/hotchemi/PermissionsDispatcher/issues/304

## Overview

- Remove optional-base plugin and bundle v13 support lib
- Update minSdkVersion to 14
- Remove bundled progurad files
- Update Gradle version to 3.3
  - remove test13 dir and move some test cases to under processor/test/java


.gitignore
```diff
@@ -2,9 +2,6 @@
 local.properties
 
 build/
-captures/
 
 .idea/
 *.iml
-
-.DS_Store
```
hotchemi: just for cleanup

.travis.yml
```diff
@@ -23,7 +23,7 @@ branches:
   except:
     - gh-pages
 
-script: ./gradlew check --stacktrace
+script: ./gradlew clean check --stacktrace
```
hotchemi: Just refactor

README.md
```diff
@@ -183,18 +183,6 @@ public class MainActivity extends AppCompatActivity {
 
 ## Note
 
-- PermissionsDispatcher depends on the `support-v4` library by default, in order to be able to use some permission compat classes.
```
hotchemi: I think these lines are no longer needed.

library/build.gradle
```diff
@@ -53,7 +51,7 @@ publish {
     groupId = GROUP_ID
     artifactId = ARTIFACT_ID_LIBRARY
     version = VERSION
-    description = DESCRIPTION
+    desc = DESCRIPTION
```
hotchemi: description became deprecated.

processor/src/main/kotlin/permissions/dispatcher/processor/util/Helpers.kt
```diff
@@ -25,7 +25,7 @@ fun permissionRequestTypeName(e: ExecutableElement): String = "${e.simpleString(
 fun <A : Annotation> findMatchingMethodForNeeds(needsElement: ExecutableElement, otherElements: List<ExecutableElement>, annotationType: Class<A>): ExecutableElement? {
     val value: List<String> = needsElement.getAnnotation(NeedsPermission::class.java).permissionValue()
     return otherElements.firstOrNull {
-        it.getAnnotation(annotationType).permissionValue().equals(value)
+        it.getAnnotation(annotationType).permissionValue() == value
```
hotchemi: Just refactor

processor/src/main/kotlin/permissions/dispatcher/processor/util/Validators.kt
```diff
@@ -38,7 +38,7 @@ fun <A : Annotation> checkDuplicatedValue(items: List<ExecutableElement>, annota
         val permissionValue = it.getAnnotation(annotationClass).permissionValue()
         Collections.sort(permissionValue)
         allItems.forEach { oldItem ->
-            if (oldItem.equals(permissionValue)) {
+            if (oldItem == permissionValue) {
```
hotchemi: Just refactor

processor/src/testV13/java/permissions/dispatcher/processor/V13ProcessorTestSuite.java
```diff
@@ -1,104 +0,0 @@
-package permissions.dispatcher.processor;
-
-import org.junit.Test;
-import permissions.dispatcher.processor.base.TestSuite;
-import permissions.dispatcher.processor.data.V13Source;
-
-public class V13ProcessorTestSuite extends TestSuite {
```
hotchemi: This class has not been executed for a long time...OMG 🙀

aurae:
In general, I approve the changes you made (raising minSdkVersion, Kotlin clean-up). However, I see that you removed the safeguards around `FragmentCompat`. As stated [here](https://github.com/hotchemi/PermissionsDispatcher/issues/304#issuecomment-298185071), this will cause an issue, because we can't really make usage of `support-v13` optional if the pre-requisite checks and `V13Access` are removed.

I had to pull the branch locally and test this out, but I was able to confirm this somewhat. Here's the example:

```groovy
// Abbreviated; I don't want native fragments, so I exclude the dependency
compile ("permissions-dispatcher:2.4.0") {
    exclude module: "support-v13"
}
annotationProcessor "..."
```

```java
@RuntimePermissions
class NativeFragment extends android.app.Fragment {
    // In this example, I'm an incosistent user that still annotates a native fragment
}
```

At runtime, the first call to `NativeFragment#xxxWithCheck()` will crash with a `ClassNotFoundException` here, since `FragmentCompat` is missing. Before, the `V13Access` class and `NativeFragmentProcessorUnit` made sure to check if it's there before trying anything.

Ideally, the annotation processor is aware of this inconsistency (but that probably won't work, since we can't rely on `FragmentCompat` on the classpath, which is the reason that issues like #279 even exist). Otherwise, maybe we can use Lint here! A Lint Detector should be able to discover the inconsistency, where I cannot annotate a native Fragment with `@RuntimePermissions`, if I don't have `FragmentCompat` on the classpath!

As an alternative, we can just remove the whole "optional dependency" thing and always use support-v13. It's only like 300 methods anyway, and ProGuard will take care of it if users don't use it. ¯\_(ツ)_/¯

library/build.gradle
aurae: Good to know! I have to check that in my other libraries :grin:

hotchemi:
@aurae Ah thx for your good catch!
So how about just keep `checkPrerequisites`? With that we'll be able to find the problem in a compile time. Adding lint check would be fine as well though!
https://github.com/hotchemi/PermissionsDispatcher/pull/306/files#diff-de57b3433a39c68273e05dc9453aaf58L18

aurae:
I'm afraid that we're going to get into a situation similar to #279 with `checkPrerequisites`... the annotation processor doesn't seem to share the entire classpath reliably with the application anymore. Therefore, we would not be able to detect the missing `FragmentCompat`, I think. Maybe a Lint check is the way to go (or we just force support-v13, that's definitely the easier option)...

hotchemi:
Okay, so let me merge this PR and create lint later! I'll address https://github.com/hotchemi/PermissionsDispatcher/issues/305 at first!

Issue #315, opened 2017-05-09
Error:Module 'com.github.hotchemi:permissionsdispatcher:2.4.0' depends on one or more Android Libraries but is a jar


zhaocheng19940105:
Error:Module 'com.github.hotchemi:permissionsdispatcher:2.4.0' depends on one or more Android Libraries but is a jar 
     compile 'com.github.hotchemi:permissionsdispatcher:2.4.0'
    annotationProcessor 'com.github.hotchemi:permissionsdispatcher-processor:2.4.0'

aurae:
Are you using Jitpack dependencies? If so, it probably tries to resolve the dependency on PD before `jcenter()` can. Since the repositories available to projects are checked in order, please move Jitpack's repo below jcenter to get rid of this issue.

@hotchemi @shiraji We should consider switching the artifact's group to avoid clashing with Jitpack.

hotchemi:
@aurae @shiraji Thx! not sure it's possible to switching the artifact's group though🤔

PR #316, opened 2017-05-09
Added a notice for Jitpack users to the README

aurae:
* related to #315, Jitpack causes an error when it tries to resolve PD's dependency itself
* users need to change the order of repositories available to their project, so that jcenter takes precedence


hotchemi:
LGTM 💯

PR #321, opened 2017-05-25
added mitene to users.md

punchdrunker:
We are using PermissionsDispatcher in our app mitene. Thank you for great lib.


hotchemi:
Brilliant💃

PR #329, opened 2017-06-21
added instructions for Kotlin

horse315:
Added instructions for new users Kotlin. If I had had it yesterday, it would have saved me about three hours.


aurae:
LGTM! @hotchemi-san, @shiraji-san, what do you think?

hotchemi:
Thank you for your suggestion, but unfortunately I don't prefer to put it that info on our README since that's apparently the common thing to use annotation processor with Kotlin and we already have a link to known issues.
So let me close that but we really appreciate for your dedicated work and will merge sample prj PR! 🙇

PR #338, opened 2017-07-07
Make lint works again!

shiraji:
Current sample project disable lint checks. Since we have some lint
checks in library, we should enable lint checks to enhance developers
use lint checks.


hotchemi:
👀

hotchemi:
great!

PR #342, opened 2017-07-12
Allow for Deterministic, Reproducible Builds with sorted inputs

### LINKED ISSUE #318: Reproducible builds
It seems that the permssionsdispatcher is breaking reproducibility of apps:

https://verification.f-droid.org/org.ligi.passandroid_338.apk.diffoscope.html

Would it be possible to always create the same code?

aurae:
The annotation processor uses a RequestCodeProvider to assign unique
request codes to each annotated @NeedsPermission method. This shared
state collides with the non-deterministic nature of the processor's
RoundEnvironment, since we can't guarantee the processing order by default.

Here, we sort the affected sets of data before starting the iterations.
First, the annotated @RuntimePermissions classes need to be in the same
order every time we build. Second, inside each annotated class, the
annotated @NeedsPermission methods also need to be in the same order.
Using simple "sortedBy" predicates, this is achieved.

fixes #318 

---

Im unsure if it's possible to unit test this. We could implement JUnit's `@Repeat` with a less trivial set of inputs, and check that the output is the same every time after n iterations. I'm unconvinced that this would greatly improve the quality of our product, though, so I'd suggest we leave it be.

### DIFF
--- a/processor/src/main/kotlin/permissions/dispatcher/processor/PermissionsProcessor.kt
+++ b/processor/src/main/kotlin/permissions/dispatcher/processor/PermissionsProcessor.kt
@@ -59,17 +59,21 @@ class PermissionsProcessor : AbstractProcessor() {
         // Create a RequestCodeProvider which guarantees unique request codes for each permission request
         val requestCodeProvider = RequestCodeProvider()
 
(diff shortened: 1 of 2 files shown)


hotchemi:
Thx! it's not a big deal basically so I'm for your thought!
LGTM

Issue #327, opened 2017-06-20
Kotlin. Internal WithCheck method generation


horse315:
Hi!
After I migrated my project to Kotlin, ```internal``` methods with ```@NeedsPermission``` annotation are generated with internal Kotlin names and look like this:
```
MainActivity2PermissionsDispatcher.`showCamera$sample_kotlin_debugWithCheck`(this)

MainActivity2PermissionsDispatcher.`showContacts$sample_kotlin_debugWithCheck`(this)
```
I had to make all these methods public.
Do you have a better solution?

horse315:
Also, ```PermissionsDispatcher``` class is not generated at the right time and module build fails. I had to comment all ```PermissionsDispatcher``` references, build the module and uncomment them back. Such behavior brakes straight CI module build.
I've prepared a sample to investigate: ```sample-kotlin``` https://github.com/horse315/PermissionsDispatcher

hotchemi:
@horse315 thx for your adopting!
Actually that's one of our [known issue](https://github.com/hotchemi/PermissionsDispatcher/blob/master/doc/known_issues.md), Kotlin `internal` keyword bundles weird info to annotation processing env so that method name becomes weird as well.
The most approachable way for now is just removing `internal` keyword from `@NeedsPermission` method. We'll plan to deal with the issue but not decided where to accomplish 🙇

ref: https://github.com/hotchemi/PermissionsDispatcher/issues/171

horse315:
@hotchemi any ideas about kotlin module build order? How can I bump annotation processor?

hotchemi:
@horse315 No problem! We'll address this problem in the near future, hopefully:D

hotchemi:
This issue will be resolved with #320 .

hotchemi:
@horse315 now you can test with `3.0.0-SNAPSHOT`!

horse315:
@hotchemi, it works great, thanks!

Issue #320, opened 2017-05-18
Kotlin based dispatcher file generation support


hotchemi:
Now we have kotlinpoet, it's surely possible to support Kotlin user by generating `***PermissionsDispatcher.kt`! I hope it would solve https://github.com/hotchemi/PermissionsDispatcher/issues/171.

https://github.com/square/kotlinpoet

aurae:
This is going to be pretty involved, but awesome if we can pull it off. There's a lot of things we can leverage with Kotlin as the target of our code generation, so I think it wouldn't necessarily be a direct port of the Java version. For example, we'll probably add the "with check" methods as extension functions to the annotated classes. In the end, we'd ideally have Java and Kotlin targets for the actual file writing, but they share the common logic of the annotation processor that we have in place today.

aurae:
What would be some conveniences that Kotlin can bring to improve the generated code, beyond supporting `internal` in a more appropriate way?

The way I see it, and what I've said before, I'd love to leverage the power of extension functions:

```kotlin
MainActivity.kt:

@RuntimePermissions
class MainActivity : AppCompatActivity() {
    @NeedsPermission(Manifest.permission.CAMERA)
    internal fun accessCamera() {
    }

    fun otherMethod() {
        // Java: "MainActivityPermissionsDispatcher.accessCameraWithCheck(this);"
        this.accessCameraWithCheck()
    }

    override fun onRequestPermissionsResult(rc: Int, permissions: Array<out String>, results: IntArray) {
        // Java: "MainActivityPermissionsDispatcher.onRequestPermissionsResult(this, rc, results);"
        // I'm not satisfied with the method signature here, since it's too similar to the Android one.
        // However, the signature is already pretty long, so I'm open for ideas.
        this.onRequestPermissionsResult(rc, results)
    }
}
```

```kotlin
MainActivityPermissionsDispatcher.kt:

fun MainActivity.accessCamera() {
    // ...
}

fun MainActivity.onRequestPermissionsResult(rc: Int, results: IntArray) {
    // ...
}
```

hotchemi:
Ah cool! I haven't thought about extension functions but it's really neat and make code base intuitive!

hotchemi:
One more thing I just came up with is coroutine. Probably we'll be able to use it for handling async suff!

https://github.com/k-kagurazaka/async-permissions

aurae:
[...]

PR #344, opened 2017-07-18
Support .kt file generation

hotchemi:
## Issues

- resolves https://github.com/hotchemi/PermissionsDispatcher/issues/320
- also resolves https://github.com/hotchemi/PermissionsDispatcher/issues/327

## Overview

- Add Kotlin file generation support.
  - Add KotlinPoet, `permissions.dispatcher.processor.impl.kotlin` classes to generate .kt file at compile time.
  - add `kotlin` method to `RuntimePermissions` to specify generate .kt file or not.
    - The reason of it is I couldn't find a way to detect file type from annotation processing info.
    - Plus, this feature is still kind of `experimental` so I'd like to allow user to generate .java file even targeted file is Kotlin.
- Address internal method
  - It seems we can address by 3c0e4fabe8a5bf386234ff45323abc292465a916...really? 😅
- Update `sample-kotlin` to use new API!

## Misc

- It turned out that it's sort of tough to support both Java/Kotlin than I expected.
  - But not sure how we can cope with it 🤔


processor/src/main/kotlin/permissions/dispatcher/processor/exception/SupportV13MissingException.kt
```diff
@@ -1,5 +0,0 @@
-package permissions.dispatcher.processor.exception
```
hotchemi: no longer used.

gradle.properties
```diff
@@ -5,21 +5,23 @@ USER = hotchemi
 GROUP_ID = com.github.hotchemi
 ARTIFACT_ID_LIBRARY = permissionsdispatcher
 ARTIFACT_ID_PROCESSOR = permissionsdispatcher-processor
-VERSION = 2.5.0-SNAPSHOT
+VERSION = 3.0.0-SNAPSHOT
```
hotchemi: 🗡

aurae: WHOOO!

hotchemi:
Not a regular work but I published 3.0.0-SNPSHOT to jfrog to prepare for today's talk at local meetup 🙇

shiraji:
Good luck! 🤞

hotchemi:
I suppose, we need to fix `test` and `test-v13` project to use generated .kt files right? 👀

hotchemi:
About https://github.com/hotchemi/PermissionsDispatcher/pull/344#issuecomment-316297145, I'll create another PR! here we go!

shiraji:
> I suppose, we need to fix test and test-v13 project to use generated .kt files right? 👀

As far as I know, yes!

PR #365, opened 2017-09-11
Rename withCheck to withPermissionCheck

hotchemi:
## Issue

addresses https://github.com/permissions-dispatcher/PermissionsDispatcher/issues/355

## Overview

- Rename withCheck to withPermissionCheck!


hotchemi:
here we go!

PR #371, opened 2017-09-15
Misc

hotchemi:
## Overview

- Remove JVM name annotation from generated file
- Update lint to alpha9.
- Replace compile to implementation, testCompiler to testImplementation.
- Apply formatter.


processor/src/main/kotlin/permissions/dispatcher/processor/impl/kotlin/KotlinBaseProcessorUnit.kt
```diff
@@ -56,7 +56,6 @@ abstract class KotlinBaseProcessorUnit : KtProcessorUnit {
     override fun createFile(rpe: RuntimePermissionsElement, requestCodeProvider: RequestCodeProvider): KotlinFile {
         return KotlinFile.builder(rpe.packageName, rpe.generatedClassName)
                 .addFileComment(FILE_COMMENT)
-                .addFileAnnotation(createFileAnnotation(rpe))
```
shiraji: We no longer use this annotation?

hotchemi: I think so! Because now generated file is only for kotlin so we don't have to consider method calling from Java side. @aurae Am I correct?

aurae: I suppose the annotation doesn't need to be there, since it's unlikely that we are going to have inter-op calls from Java into the generated Kotlin code, which is tightly bound to the target also available through Kotlin.

shiraji: Got it!

Issue #376, opened 2017-09-17
NeedsPermission annotated method with parameter doesn't work in Kotlin


hotchemi:
Log: https://gist.github.com/ryugoo/cf8d4057973eac940c7713e054c02c35

shiraji:
There are multiple issues here.
* kotlin.String converts to java.lang.String
* `GrantableRequest` declares as `val`
* `GrantableRequest` should be nullable and initializes as null
* `XxxPermissionRequest` has property that is the same name with primary constructor's one.

I am working on GrantableRequest and XxxPermissionRequest problems.

hotchemi:
Thx! I've just worked on it and here's a patch file.

[patch.txt](https://github.com/permissions-dispatcher/PermissionsDispatcher/files/1308942/patch.txt)

PR #380, opened 2017-09-17
Fix syntax error when param exists

shiraji:
Related with #376

Haven't resolve the problem with kotlin.String parameter.

### DIFF
--- a/processor/src/main/kotlin/permissions/dispatcher/processor/util/Helpers.kt
+++ b/processor/src/main/kotlin/permissions/dispatcher/processor/util/Helpers.kt
@@ -59,7 +59,7 @@ fun varargsParametersCodeBlock(needsElement: ExecutableElement): CodeBlock {
 fun varargsKtParametersCodeBlock(needsElement: ExecutableElement): com.squareup.kotlinpoet.CodeBlock {
     val varargsCall = com.squareup.kotlinpoet.CodeBlock.builder()
     needsElement.parameters.forEachIndexed { i, it ->
-        varargsCall.add("\$L", it.simpleString())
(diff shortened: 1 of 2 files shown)


processor/src/main/kotlin/permissions/dispatcher/processor/util/Helpers.kt
```diff
@@ -48,7 +48,7 @@ fun <A : Annotation> findMatchingMethodForNeeds(needsElement: ExecutableElement,
 fun varargsParametersCodeBlock(needsElement: ExecutableElement): CodeBlock {
     val varargsCall = CodeBlock.builder()
     needsElement.parameters.forEachIndexed { i, it ->
-        varargsCall.add("\$L", it.simpleString())
```
hotchemi: In JavaPoet this writing style is appropriate. This is actually super misleading though

sample-kotlin/build.gradle
```diff
@@ -27,5 +27,6 @@ dependencies {
     implementation (project(':library')) {
         exclude module: "support-v13"
     }
+    provided project(':processor')
```
hotchemi: Oh I thought this causes compile error with the latest android gradle plugin 👀

shiraji: You are right. I need to remove this one.

processor/src/main/kotlin/permissions/dispatcher/processor/util/Helpers.kt
shiraji: Yes

hotchemi:
Thank you! LGTM!

Issue #392, opened 2017-10-08
Lint could not load custom rule jar with android gradle plugin v2.3.3


OsipXD:
## Overview

- Can't build project after add PermissionDispatcher

### Expected

- Project building should be successful

### Actual

- I have error that looks like conflict with lint and crashes my build
```java
Could not load custom rule jar file /home/user/.android/build-cache/317bbf7182688f2c4e85a24d078fe2ab62ffb2a0/output/jars/lint.jar
java.lang.NoClassDefFoundError: com/android/tools/lint/detector/api/Detector$UastScanner
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:763)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:467)
	at java.net.URLClassLoader.access$100(URLClassLoader.java:73)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:368)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:362)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:361)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at permissions.dispatcher.PermissionsDispatcherIssueRegistry.getIssues(PermissionsDispatcherIssueRegistry.java:13)
	at com.android.tools.lint.client.api.JarFileIssueRegistry.<init>(JarFileIssueRegistry.java:125)
	at com.android.tools.lint.client.api.JarFileIssueRegistry.get(JarFileIssueRegistry.java:90)
	at com.android.tools.lint.client.api.LintDriver.registerCustomDetectors(LintDriver.java:603)
	at com.android.tools.lint.client.api.LintDriver.analyze(LintDriver.java:521)
	at com.android.tools.lint.client.api.LintDriver.analyze(LintDriver.java:483)
	at com.android.tools.lint.LintCliClient.run(LintCliClient.java:143)
	at com.android.build.gradle.internal.LintGradleClient.run(LintGradleClient.java:197)
	at com.android.build.gradle.tasks.Lint.runLint(Lint.java:340)
	at com.android.build.gradle.tasks.Lint.lintAllVariants(Lint.java:140)
[...]

PR #393, opened 2017-10-11
Combine Uast & Psi Lint rules

mannodermaus:
Our `IssueRegistry` now operates on a List of "Provider" implementations,
which serve as candidates when it comes to providing a list of Issues
to detect via Lint. By default, `UastScanner` implementations are queried
first, and a fallback to the old `JavaPsiScanner` classes is made when
`UastScanner` is absent on the classpath.

The unit tests for the `IssueRegistry` have been extended to account for
changing Provider implementations with varying availability. Also, issue
creation has been moved into a new class `IssueFactory` to avoid
redundant re-definition of Issues for each Provider.

This is related to #392, where an older Lint runtime without `UastScanner`
can't invoke our custom Lint rules. For those cases, we now fall back to Psi.

I would love to have some sort of "logging" inside the `IssueRegistry`, so
we can make sure to get more info on the environment of future Lint issues
(i.e. "is that a bug in our Uast or Psi rules?"). Since I'm unsure how to gain
access to the `LintClient` from inside an `IssueRegistry`, this is done
using simple `System.out` statements at the moment.


lint/build.gradle
```diff
@@ -24,3 +24,14 @@ jar {
...
     }
 }
+
+test {
+    // Ensure that ANDROID_HOME is set when launching Lint tests
+    // (non-standard shells won't fill the reference, leading to
+    // "This test requires an Android SDK" errors at runtime)
+    if (System.getenv("ANDROID_HOME") == null) {
+        Properties localProps = new Properties()
+        file("$rootDir/local.properties").withReader { localProps.load(it) }
```
mannodermaus: Now that I think about it, this would be prettier:

```groovy
rootProject.file("local.properties").withReader { localProps.load(it) }
```

lint/src/test/java/permissions/dispatcher/Utils.java
```diff
@@ -2,15 +2,16 @@
 
 import org.intellij.lang.annotations.Language;
 
-final class Utils {
+public final class Utils {
```
mannodermaus: I used to have different packages for `psi` and `uast` implementations, which made the visibility change of this class necessary. Now that everything ended up in 1 package after all, I will probably revert this back.

lint/src/main/java/permissions/dispatcher/detectors/PsiCallNeedsPermissionDetector.java
```diff
@@ -0,0 +1,117 @@
...
+import com.intellij.psi.PsiReferenceExpression;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.EnumSet;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+public class PsiCallNeedsPermissionDetector extends Detector implements Detector.JavaPsiScanner {
```
hotchemi: presumably current implementation doesn't pass `UastCallNeedsPermissionDetectorTest`, I found there's a logic mistake on https://github.com/permissions-dispatcher/PermissionsDispatcher/pull/368 so we need to align the logic of `PsiCallNeedsPermissionDetector` to `UastCallNeedsPermissionDetector` 😅

lint/src/main/java/permissions/dispatcher/PermissionsDispatcherIssueRegistry.java
```diff
@@ -4,15 +4,101 @@
...
+        this.providers = providers;
+    }
+
     @Override
     public List<Issue> getIssues() {
-        return Arrays.asList(
-                CallNeedsPermissionDetector.ISSUE,
-                CallOnRequestPermissionsResultDetector.ISSUE,
-                NoCorrespondingNeedsPermissionDetector.ISSUE);
+        for (Provider provider : providers) {
```
hotchemi: Nit: for me those lines look a bit awkward because it depends on the order of `DEFAULT_PROVIDERS` and if we change the order it would get messed up right? just checking classpath at first is enough for me👀

mannodermaus: Yes, `DEFAULT_PROVIDERS` is an ordered List where each element is checked in sequence! There is a comment notifying about the relevance & implications of that at the declaration site of that variable, though.

lint/src/main/java/permissions/dispatcher/detectors/PsiCallNeedsPermissionDetector.java
mannodermaus: Interesting! Let's migrate the test to #368's version, then compare the results again. :)

hotchemi:
Any update? 🙇 Thx!

mannodermaus:
Terribly sorry about the delay, I haven't had much time to dedicate to this PR in the last couple of days. I intend to pick this up again very soon, though! 🙇

mannodermaus:
I have restructured our Lint test cases so that both Uast & Psi Detectors run against them. This should theoretically help us keep the different implementations in sync. What do you guys think? Also, @hotchemi I haven't had the time to deeper investigate the logic changes you made in #368, so I'm unsure what to change here. If you don't mind taking over, you can go ahead! The unit test basis is laid out now :).

hotchemi:
@mannodermaus sure! can you push your branch to this repo and recreate the PR? I'd like to hand over.

mannodermaus:
Continued in #400 by request

PR #400, opened 2017-10-25
Combine Uast & Psi Lint rules

mannodermaus:
Continuation of #393


mannodermaus:
With the stable release of AS 3.0 and the accompanying Gradle plugin, I'd advocate for actually dropping this two-way implementation to remain backwards compatible. What do you guys think?

hotchemi:
Yeah I agree with you! AS3 is **stable** now so we don't have to add a back compatibility I guess 🙇  Thank you for your dedicated work!

PR #401, opened 2017-10-26
Replace local Android SDK dependency for processor unit tests

mannodermaus:
For ages, we included a dependency on [com.google.android:android](https://mvnrepository.com/artifact/com.google.android/android) in order to allow our processor's unit tests to run properly with the library's Android-based code. Because that dependency hasn't been updated for like two years now, important build constants like `Build.VERSION_CODES.M` were missing from that version. By extension, we weren't allowed to reference code like that directly, leading to awkward workarounds and comments explaining why our code can't be cleaner.

This PR gets rid of this nasty dependency, and replaces it with a lightweight custom Gradle plugin, which allows us to depend on the local SDK installation's `android.jar` instead - i.e. the proper way of depending on Android SDK things locally. Whenever we update the Android compile versions, the new APIs can now be safely referenced in our library code, and the processor's tests will still be able to compile properly.


hotchemi:
👁

hotchemi:
Sorry for the late, this is awesome 😄 
Just curious, is it possible to remove these jars as well? 👀 https://github.com/permissions-dispatcher/PermissionsDispatcher/tree/master/processor/src/test/libs

mannodermaus:
I'll give it some thought, but it might not be as easy, since the support library is served through a "real" Maven repo. Then again, I've read some interesting approach to this in Robolectric's code, maybe we can do something with that.

hotchemi:
Ah ic! So let me merge the PR for the time being~

mannodermaus:
As a follow-up to the question in #403: I haven't been able to migrate Groovy's dynamic `ExtensionAware` syntax to Kotlin so far, which is why I didn't find a way to create the dependency handler yet:

```groovy
project.dependencies.ext.androidJar = {
  // ...
}
```

This doesn't have an equivalent in Kotlin because of its strongly-typed nature. You can have a go at this yourself if you like, but I'd say that this plugin needs to stay in Groovy for now.

PR #403, opened 2017-11-01
Add AAR-to-JAR plugin to decouple processor's test scope from Android

mannodermaus:
This is a continuation of #401, in which it was suggested that we might be able to clean our dependency graphs even further. This is the proposal to get rid of our local JAR files.

___

The annotation processor required knowledge of Android classes
in order to perform its unit tests for code generation. Therefore,
we used to have two JAR versions of `support-v4` & `support-v13` in its
`src/test/libs` folder, which haven't been updated in quite some time.
Additionally, we had to mock out `android.provider.Settings` manually,
just to make the test scope compile, since we depended on the `library`
project directly through declaration in the `processor`'s build script
(`sourceSets.test.java.srcDir`).

With this commit, these archaic discrepancies are laid to rest, and
replaced with a cleaner separation between Java and Android for our
processor. No dependencies are leaking into our compilation classpath
anymore, so the processor doesn't have to depend on the library's Android
stuff explicitly anymore. Instead, we utilize a new Gradle plugin which
generates "JAR stubs" from AAR libraries to provide a mere facade for
code generation tests.

There's two parts to this: First, a new plugin inside our `buildSrc`
project takes care of converting AAR dependencies of the `processor` project
to JAR, using a custom configuration called `testCompileAar`. It also
records the location of the converted JAR stubs into a resource file
accessible by the processor's test scope. Secondly, this resource file
is used to construct an enhanced classpath which is "Android-aware", then
provided to our code generation tests in `TestSuite.java`.

This will also pave the way to finally rid ourselves of the two source
directories inside the library projects. There's nothing really stopping
us from putting all the annotations inside their own module now, leaving
only `PermissionUtils.java` in the `library` project, and removing the
additional `srcDir` declaration from the `processor` project. This would be
replacible by a dependency on the `annotations` project!


hotchemi:
Thank you! This PR is actually amazing!
My only question is do we need to write the plugin with Groovy? 😁

mannodermaus:
We don't use anything _too_ Groovy-specific in terms of how the code is written, so Kotlin should theoretically be possible. It gets iffy when you utilize Groovy's dynamic extensions (i.e. `project.ext.something`), since those aren't really exposed as methods, and interoperating between Kotlin and `GObject` didn't really work that well when I tried it last time. I'll give it some thought

hotchemi:
Thx! it's not urgent so let me merge~

mannodermaus:
I have successfully converted this plugin to Kotlin. Do you want me to open a PR? The other plugin described in #401 wasn't so lucky... see my comment over there for details.

PR #410, opened 2017-12-02
Fix problem with special permission

### LINKED ISSUE #406: PD does not generate kt version of SYSTEM_ALERT_WINDOW
PD version is 3.0.1

if you set `NeedsPermission(SYSTEM_ALERT_WINDOW)`, it won't generate code.

This is because `SystemAlertWindowHelper` use JavaPoet's ClassName.

https://github.com/permissions-dispatcher/PermissionsDispatcher/blob/master/processor/src/main/kotlin/permissions/dispatcher/processor/impl/kotlin/SystemAlertWindowHelper.kt#L3

Also, I guess `onActivityResult` method should be extension method of Activity.

I will fix these and add test cases.

shiraji:
Close #406

### DIFF
--- a/test/src/test/java/permissions/dispatcher/test/ActivityWithSystemAlertWindowAllAnnotationsPermissionsDispatcherTest.kt
+++ b/test/src/test/java/permissions/dispatcher/test/ActivityWithSystemAlertWindowAllAnnotationsPermissionsDispatcherTest.kt
@@ -1,0 +1,156 @@
+package permissions.dispatcher.test
+
+import android.content.Intent
+import android.net.Uri
+import android.os.Build
+import android.os.Process
+import android.provider.Settings
+import android.support.v4.app.ActivityCompat
+import android.support.v4.app.AppOpsManagerCompat
+import android.support.v4.content.PermissionChecker
+import org.junit.Before
+import org.junit.BeforeClass
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mockito.Matchers
+import org.mockito.Mockito
+import org.powermock.api.mockito.PowerMockito
+import org.powermock.core.classloader.annotations.PrepareForTest
+import org.powermock.modules.junit4.PowerMockRunner
+import permissions.dispatcher.test.ActivityWithSystemAlertWindowAllAnnotationsPermissionsDispatcher.*
+import permissions.dispatcher.PermissionRequest
+
+@Suppress("IllegalIdentifier")
+@RunWith(PowerMockRunner::class)
+@PrepareForTest(ActivityCompat::class, PermissionChecker::class,
+        AppOpsManagerCompat::class, Process::class, Settings::class, Build.VERSION::class, Uri::class)
+class ActivityWithSystemAlertWindowAllAnnotationsPermissionsDispatcherTest {
+    private lateinit var activity: ActivityWithSystemAlertWindowAllAnnotations
+
+    companion object {
+        private var requestCode = 0
+
+        @BeforeClass
+        @JvmStatic
+        fun setUpForClass() {
+            // TODO Reflection on Kotlin top-level properties?
+            requestCode = 7
+        }
+    }
+
+    @Before
+    fun setUp() {
+        activity = Mockito.mock(ActivityWithSystemAlertWindowAllAnnotations::class.java)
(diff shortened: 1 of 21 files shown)


test/src/test/java/permissions/dispatcher/test/ActivityWithSystemAlertWindowAllAnnotationsPermissionsDispatcherTest.kt
```diff
@@ -0,0 +1,156 @@
...
+    private lateinit var activity: ActivityWithSystemAlertWindowAllAnnotations
+
+    companion object {
+        private var requestCode = 0
+
+        @BeforeClass
+        @JvmStatic
+        fun setUpForClass() {
+            // TODO Reflection on Kotlin top-level properties?
+            requestCode = 7
```
shiraji: I really don't like this code but I'm not sure how to get private top level properties using kotlin-reflection...

hotchemi:
👀

test/src/test/java/permissions/dispatcher/test/ActivityWithSystemAlertWindowAllAnnotationsPermissionsDispatcherTest.kt
mannodermaus: Yeah, I had the same problem once, no idea.

mannodermaus:
👀

test/src/test/java/permissions/dispatcher/test/ActivityWithSystemAlertWindowAllAnnotationsPermissionsDispatcherTest.kt
shiraji: I asked the question to reflect channel in Kotlinlang

https://kotlinlang.slack.com/archives/C0AVAB92L/p1512281918000028

let's see

mannodermaus: TIL of the `#reflect` channel on kotlinlang's Slack. Nice!

shiraji: well, I think at this moment there is no way to solve this problem...

hotchemi:
Ah sorry I'll check again🙇

test/src/test/java/permissions/dispatcher/test/ActivityWithSystemAlertWindowAllAnnotationsPermissionsDispatcherTest.kt
mannodermaus: Did you open an issue on YouTrack, by the way? If so, could you share the URL?

shiraji:
I'm not sure why CI failed, thou

test/src/test/java/permissions/dispatcher/test/ActivityWithSystemAlertWindowAllAnnotationsPermissionsDispatcherTest.kt
shiraji: Not yet, but I just notice I don't even know how to access a private top property. I read some article that says there is no way to access private top property. I couldn't find it at this moment.

...well, let me open the issue.

shiraji: https://youtrack.jetbrains.com/issue/KT-21707

shiraji: Sadly, 

```
there is no way to access private top property
```

this statement is correct... 😢 

https://youtrack.jetbrains.com/issue/KT-21707#comment=27-2595536

hotchemi:
Umm...I tried to rebuild and it passed...may be we need to reset something after test passed? 🤔
Anyway, let me merge and investigate the problem separately~! Thanks!

hotchemi:
Ah ok may be test being failed when multiple processes execute test simultaneously... 🤔

PR #438, opened 2018-02-27
Update KotlinPoet to 0.8.0-SNAPSHOT

hotchemi:
To avoid issues/437 bug.
See https://github.com/permissions-dispatcher/PermissionsDispatcher/issues/437.


hotchemi:
let me close for the time being.

PR #456, opened 2018-04-11
[WIP] Make CI stable

hotchemi:


mannodermaus:
Interesting how it won't pass on OpenJDK, but it does on Oracle, hm...

shiraji:
It seems this method sometimes work, sometimes not work.
https://github.com/permissions-dispatcher/PermissionsDispatcher/blob/b8e0bedada1714b624d0d7ad729b7d94dd1140d1/test/src/test/java/permissions/dispatcher/test/Extensions.kt#L34-L37

I'm not sure but `PowerMockito` or this static method is affected from other test cases/test case order?

hotchemi:
Actually the test cases can be failed on both JDK, but never fail on my local..🤔
I'll try to use CirclCI just in case.

PR #458, opened 2018-04-14
Add JvmName annotation to generated file

WonderCsabo:
This way Java code can simply call the generated methods,
without appending `Kt` to end of the generated class name.

This should fix https://github.com/permissions-dispatcher/AndroidAnnotationsPermissionsDispatcherPlugin/issues/15 .


hotchemi:
LGTM 💯

PR #459, opened 2018-04-14
Fix intermittent failure of test project

hotchemi:
## Issue

resolves https://github.com/permissions-dispatcher/PermissionsDispatcher/issues/455

## Overview

Inspired by eclipse/hono#145.

occasionally we've suffered `NeverWantedButInvoked` from mockito only on TravisCI.. 🤔
I've tried many points and finally avoiding keeping activity/fragment as a member variable somehow address the problem..I tried building 10 times on TravisCi but seemingly never fails.. 😅


shiraji:
LGTM (Doesn't make sense to me but it works)

hotchemi:
yea it's weird..

mannodermaus:
wtf, that really is the weirdest thing. But oh well, if it's stupid and it works, it's not stupid 🤷‍♂️

PR #475, opened 2018-05-19
Separate annotations to dedicated module

hotchemi:


library/build.gradle
```diff
@@ -22,7 +22,6 @@ android {
         main {
             java {
                 srcDirs = [
-                        'src/main/java',
                         'src/main/android'
```
mannodermaus: Can we move the entire source to`src/main/java` now?

annotation/build.gradle
```diff
@@ -0,0 +1,4 @@
+apply plugin: 'java-library'
```
mannodermaus: Don't we need Maven coordinates & publish this to jcenter as well?

library/build.gradle
hotchemi: thx 🙇3126812

annotation/build.gradle
hotchemi: If I'm correct we don't have to publish this module since we use [`api` configuration](https://github.com/permissions-dispatcher/PermissionsDispatcher/pull/475/files#diff-0d1d9100a9b6813b8cb5c470bf313d00R35)!
 
ref: https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_separation

mannodermaus: Ah, so it will be bundled in the library's JAR? Let's verify that with a local build

hotchemi: I guess so because sample module compiled successfully without adding anything!

PR #479, opened 2018-06-04
Fix the problem with order matter "Useless @OnShowRationale declaration"

### LINKED ISSUE #407: Lint showing "Useless @OnShowRationale declaration"
___

## Overview

- Describe the issue briefly

I updated the library to 3.0.1 from 2.4.0 and now I'm seeing a Lint error saying "Useless @OnShowRationale declaration". It's not stopping me from compiling my code but I'm wondering why it's showing up because when I remove it, the function is no longer being used. Maybe it's because I'm checking for 2 permissions in the same Fragment?

### Expected

- What is the expected behavior?

I guess not for it to show up since it doesn't seem to be useless.

### Actual

- What is the actual behavior?

## Environment

- Which library version are you using? 

3.0.1
- On which devices do you observe the issue?
- Note any other information that might be useful

## Reproducible steps

- While it's not required, it'd be perfect to add a link to a sample project where you encounter the issue

It's happening in [this class](https://github.com/silverAndroid/FriendStagram-Android/blob/master/app/src/main/java/rbsoftware/friendstagram/ui/fragment/SelectImageFragment.kt)

shiraji:
The problem is it validated each time lint found annotation. This commit
make lint validate the annotation at afterVisitClass since at the point,
all annotations are stored into needsPermissionsAnnotations and
onShowRationaleAnnotations variables.

Close #407

### DIFF
--- a/lint/src/main/java/permissions/dispatcher/NoCorrespondingNeedsPermissionDetector.java
+++ b/lint/src/main/java/permissions/dispatcher/NoCorrespondingNeedsPermissionDetector.java
@@ -75,6 +75,11 @@ public final class NoCorrespondingNeedsPermissionDetector extends Detector imple
             if (onShowRationaleAnnotations.isEmpty()) {
                 return true;
             }
+            return true;
+        }
+
+        @Override
+        public void afterVisitClass(UClass node) {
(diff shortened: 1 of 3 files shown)


hotchemi:
GO BOLD

PR #480, opened 2018-06-04
Make CI stable, again

### LINKED ISSUE #455: [Maintainers]CI is unstable
Currently TravisCI occasionally fails around PowerMock.. 🤔 JDK or environment problem? 🤔 

https://travis-ci.org/permissions-dispatcher/PermissionsDispatcher/builds/364717010

shiraji:
The problem was it set `Build.MANIFACTURE` using reflection https://github.com/permissions-dispatcher/PermissionsDispatcher/blob/980c35ed396ce7d896c69b2b862923e38b7af11f/test/src/test/java/permissions/dispatcher/test/Extensions.kt#L73-L81 and did not clear the value. So, that since `Build.MANIFACTURE` is xiaomi, https://github.com/permissions-dispatcher/PermissionsDispatcher/blob/39c665901c0b293182be8b236a1524807dbca2b1/library/src/main/java/permissions/dispatcher/PermissionUtils.java#L94-L97 `hasSelfPermissionForXiaomi` get called and the tests fail.

This happens when xiaomi related tests get executed before testing OnShowRationale related tests.

You can reproduce this issue in local machine if https://github.com/permissions-dispatcher/PermissionsDispatcher/blob/f0f55196dea473eb9dbe729f0635c4da8fde6577/test/src/test/java/permissions/dispatcher/test/ActivityOnlyNeedsPermissionPermissionsDispatcherTest.kt#L93-L102 move to top of the test methods. (I cannot use `FixMethodOrder` to force the test order to reproduce this issue because the tests uses PowerMock.)
 
Close #455

### DIFF
--- a/test-v13/build.gradle
+++ b/test-v13/build.gradle
@@ -24,6 +24,17 @@ android {
     sourceSets {
         main.java.srcDirs += "src/main/kotlin"
     }
+
+    testOptions {
+        unitTests.all {
+            testLogging {
+                events "passed", "failed", "standardOut", "standardError"
+                outputs.upToDateWhen {false}
+                showStandardStreams = true
+                exceptionFormat "full"
+            }
+        }
+    }
 }
 
 dependencies {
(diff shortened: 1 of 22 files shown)


hotchemi:
Let me check!

test-v13/build.gradle
```diff
@@ -24,6 +24,17 @@ android {
     sourceSets {
         main.java.srcDirs += "src/main/kotlin"
     }
+
+    testOptions {
```
hotchemi: Do we need this? 👁

shiraji: I put this in case this happens again. Hopefully never...

PR #482, opened 2018-06-04
Fix compile error when request SYSTEM_ALERT_WINDOW on SupportFragment

### LINKED ISSUE #429: Fragment's onActivityResult() is not called when requesting SYSTEM_ALERT_WINDOW permission
## Overview

- I'm using `@RuntimePermissions` with  an `android.support.v4.app.Fragment` and I need request `SYSTEM_ALERT_WINDOW` permission using `@NeedsPermission(Manifest.permission.SYSTEM_ALERT_WINDOW)`.
According to documentation I must override `onActivityResult()` calling the generated `onActivityResult()` of PermissionDispatcher. However, Fragment's onActivityResult() is not called when I back from app settings.
The issues occurs because the generated fragment extension use `activity.startActivityForResult()`. When using activity method, the fragment's onActivityResult() is not called. The library should use `fragment.startActivityForResult()` instead.

### Expected

- The method `onActivityResult()`of Fragment should be called.

### Actual

- The method `onActivityResult()`of Fragment is not called.

## Environment

Library version is 3.1.0
Kotlin plugin 1.1.60

tomoya0x00:
This problem seems to have happened because `android.support.v4.app.Fragment.getActivity()` is specified as nullable.

And I also fix #429.

### DIFF
--- a/processor/src/main/kotlin/permissions/dispatcher/processor/impl/kotlin/KotlinSupportFragmentProcessorUnit.kt
+++ b/processor/src/main/kotlin/permissions/dispatcher/processor/impl/kotlin/KotlinSupportFragmentProcessorUnit.kt
@@ -1,7 +1,7 @@
 package permissions.dispatcher.processor.impl.kotlin
 
 import com.squareup.kotlinpoet.FunSpec
-import permissions.dispatcher.processor.util.*
+import permissions.dispatcher.processor.util.typeMirrorOf
 import javax.annotation.processing.Messager
 import javax.lang.model.type.TypeMirror
 
@@ -14,7 +14,7 @@ class KotlinSupportFragmentProcessorUnit(messager: Messager) : KotlinBaseProcess
 
     override fun getTargetType(): TypeMirror = typeMirrorOf("android.support.v4.app.Fragment")
 
-    override fun getActivityName(): String = "activity"
+    override fun getActivityName(targetParam: String): String = "$targetParam.activity"
 
     override fun addShouldShowRequestPermissionRationaleCondition(builder: FunSpec.Builder, permissionField: String, isPositiveCondition: Boolean) {
         val condition = if (isPositiveCondition) "" else "!"
--- a/processor/src/main/kotlin/permissions/dispatcher/processor/impl/kotlin/SystemAlertWindowHelper.kt
+++ b/processor/src/main/kotlin/permissions/dispatcher/processor/impl/kotlin/SystemAlertWindowHelper.kt
@@ -14,8 +14,8 @@ class SystemAlertWindowHelper : SensitivePermissionInterface {
         builder.beginControlFlow("if (%T.hasSelfPermissions(%N, *%N) || %T.canDrawOverlays(%N))", PERMISSION_UTILS, activity, permissionField, SETTINGS, activity)
     }
 
-    override fun addRequestPermissionsStatement(builder: FunSpec.Builder, activity: String, requestCodeField: String) {
-        builder.addStatement("val intent = %T(%T.ACTION_MANAGE_OVERLAY_PERMISSION, %T.parse(\"package:\" + %N.getPackageName()))", INTENT, SETTINGS, URI, activity)
-        builder.addStatement("%N.startActivityForResult(intent, %N)", activity, requestCodeField)
+    override fun addRequestPermissionsStatement(builder: FunSpec.Builder, targetParam: String, activityVar: String, requestCodeField: String) {
+        builder.addStatement("val intent = %T(%T.ACTION_MANAGE_OVERLAY_PERMISSION, %T.parse(\"package:\" + %N!!.getPackageName()))", INTENT, SETTINGS, URI, activityVar)
+        builder.addStatement("%N.startActivityForResult(intent, %N)", targetParam, requestCodeField)
     }
 }
--- a/processor/src/main/kotlin/permissions/dispatcher/processor/impl/kotlin/KotlinActivityProcessorUnit.kt
+++ b/processor/src/main/kotlin/permissions/dispatcher/processor/impl/kotlin/KotlinActivityProcessorUnit.kt
@@ -2,7 +2,7 @@ package permissions.dispatcher.processor.impl.kotlin
 
 import com.squareup.kotlinpoet.ClassName
 import com.squareup.kotlinpoet.FunSpec
-import permissions.dispatcher.processor.util.*
+import permissions.dispatcher.processor.util.typeMirrorOf
 import javax.annotation.processing.Messager
 import javax.lang.model.type.TypeMirror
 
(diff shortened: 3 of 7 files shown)


shiraji:
Current generated code where we got syntax error for `activity.getPackageName()` and `activity.startActivityForResult(intent, REQUEST_FOO)`

```kotlin
fun ContactsFragment.fooWithPermissionCheck() {
  if (PermissionUtils.hasSelfPermissions(activity, *PERMISSION_FOO) || Settings.canDrawOverlays(activity)) {
    foo()
  } else {
    val intent = Intent(Settings.ACTION_MANAGE_OVERLAY_PERMISSION, Uri.parse("package:" + activity.getPackageName()))
    activity.startActivityForResult(intent, REQUEST_FOO)
  }
}

fun ContactsFragment.onActivityResult(requestCode: Int) {
  when (requestCode) {
    REQUEST_FOO -> {
      if (PermissionUtils.hasSelfPermissions(activity, *PERMISSION_FOO) || Settings.canDrawOverlays(activity)) {
        foo()
      }
    }
  }
}
```

Generated Code

```kotlin
fun ContactsFragment.fooWithPermissionCheck() {
  if (PermissionUtils.hasSelfPermissions(this.activity, *PERMISSION_FOO) || Settings.canDrawOverlays(this.activity)) {
    foo()
  } else {
    val intent = Intent(Settings.ACTION_MANAGE_OVERLAY_PERMISSION, Uri.parse("package:" + this.activity!!.getPackageName()))
    this.startActivityForResult(intent, REQUEST_FOO)
  }
}

fun ContactsFragment.onActivityResult(requestCode: Int) {
  when (requestCode) {
    REQUEST_FOO -> {
      if (PermissionUtils.hasSelfPermissions(this.activity, *PERMISSION_FOO) || Settings.canDrawOverlays(this.activity)) {
        foo()
      }
    }
  }
}
```

processor/src/main/kotlin/permissions/dispatcher/processor/impl/kotlin/KotlinSupportFragmentProcessorUnit.kt
```diff
@@ -14,7 +14,7 @@ class KotlinSupportFragmentProcessorUnit(messager: Messager) : KotlinBaseProcess
 
     override fun getTargetType(): TypeMirror = typeMirrorOf("android.support.v4.app.Fragment")
 
-    override fun getActivityName(): String = "activity"
+    override fun getActivityName(targetParam: String): String = "$targetParam.activity"
```
shiraji: I wonder can we ignore `targetParam` here?

```kotlin
override fun getActivityName(targetParam: String): String = "activity"
```

shiraji: The answer is No! We have `PermissionRequest` class

```kotlin
val intent = Intent(Settings.ACTION_MANAGE_OVERLAY_PERMISSION, Uri.parse("package:" + target.activity!!.getPackageName()))
```

shiraji:
We really need compiling test for kotlin now...

processor/src/main/kotlin/permissions/dispatcher/processor/impl/kotlin/SystemAlertWindowHelper.kt
```diff
@@ -14,8 +14,8 @@ class SystemAlertWindowHelper : SensitivePermissionInterface {
         builder.beginControlFlow("if (%T.hasSelfPermissions(%N, *%N) || %T.canDrawOverlays(%N))", PERMISSION_UTILS, activity, permissionField, SETTINGS, activity)
     }
 
-    override fun addRequestPermissionsStatement(builder: FunSpec.Builder, activity: String, requestCodeField: String) {
-        builder.addStatement("val intent = %T(%T.ACTION_MANAGE_OVERLAY_PERMISSION, %T.parse(\"package:\" + %N.getPackageName()))", INTENT, SETTINGS, URI, activity)
-        builder.addStatement("%N.startActivityForResult(intent, %N)", activity, requestCodeField)
+    override fun addRequestPermissionsStatement(builder: FunSpec.Builder, targetParam: String, activityVar: String, requestCodeField: String) {
+        builder.addStatement("val intent = %T(%T.ACTION_MANAGE_OVERLAY_PERMISSION, %T.parse(\"package:\" + %N!!.getPackageName()))", INTENT, SETTINGS, URI, activityVar)
```
shiraji: We can replace `!!` with `?`. Like `parse("package: " + activity?.getPackageName())` it ends up `parse("package: null")`. What do you think ? @hotchemi @tomoya0x00 

However, if we could add null check to Java version, we can do early return in kotlin so that we don't need to worry about this problem.

```kotlin
fun ContactsFragment.fooWithPermissionCheck() {
  val activity = activity ?: return // add this 
  if (PermissionUtils.hasSelfPermissions(activity, *PERMISSION_FOO) || Settings.canDrawOverlays(activity)) {
    foo()
  } else {
    val intent = Intent(Settings.ACTION_MANAGE_OVERLAY_PERMISSION, Uri.parse("package:" + activity.getPackageName()))
    activity.startActivityForResult(intent, REQUEST_FOO)
  }
}
```

shiraji: I'm not sure we should do this in this PR. We can ship this PR and then later version we can add null check in both Java and Kotlin.

tomoya0x00:
@shiraji 

> We really need compiling test for kotlin now...

I think so too...

Does compiling test for Java exist?
I would like to refer if it exists.

@hotchemi 
> is it possible to avoid using !!?

I'll try it.
If activity is null, what do you think of the idea of ​​`throw IllegalStateException` ?

processor/src/main/kotlin/permissions/dispatcher/processor/impl/kotlin/SystemAlertWindowHelper.kt
tomoya0x00: I missed this comment.

I think we can add a null check later.

I think that `early return` is better. So We already use it in PermissionRequest class, I feel a sense of unity.

shiraji:
@tomoya0x00 

> Does compiling test for Java exist?

[compile-testing](https://github.com/google/compile-testing)

There is no kotlin compiler testing that tests kapt generated test. @mannodermaus asks moshi team to open kotlin code gen test class

https://github.com/square/moshi/blob/3ecdfb6374758678859d1a75cb1b890ac28656fa/kotlin/codegen/src/test/java/com/squareup/moshi/kotlin/codegen/JsonClassCodegenProcessorTest.kt

processor/src/main/kotlin/permissions/dispatcher/processor/impl/kotlin/SystemAlertWindowHelper.kt
hotchemi: Yea I definitely prefer the latter one. But it can be another PR for sure~

hotchemi:
OK let me merge! Hopefully someone would address `early return`..? cc: @tomoya0x00 😁

Issue #483, opened 2018-06-06
Handling support fragment's getActivity()


shiraji:
Currently, we ignore the case when `getActivity()` return `null`

```kotlin
fun ContactsFragment.fooWithPermissionCheck() {
  if (PermissionUtils.hasSelfPermissions(activity, *PERMISSION_FOO) || Settings.canDrawOverlays(activity)) {
    foo()
  } else {
    val intent = Intent(Settings.ACTION_MANAGE_OVERLAY_PERMISSION, Uri.parse("package:" + activity!!.getPackageName())) // NPE if activity is null
    activity.startActivityForResult(intent, REQUEST_FOO)
  }
}
```

We should take care when `getActivity()` return `null`. The following code is just one idea I came up.

```kotlin
fun ContactsFragment.fooWithPermissionCheck() {
  val activity = activity ?: return // add this 
  if (PermissionUtils.hasSelfPermissions(activity, *PERMISSION_FOO) || Settings.canDrawOverlays(activity)) {
    foo()
  } else {
    val intent = Intent(Settings.ACTION_MANAGE_OVERLAY_PERMISSION, Uri.parse("package:" + activity.getPackageName()))
    activity.startActivityForResult(intent, REQUEST_FOO)
  }
}
```

See #482

mannodermaus:
Are we only going to add the check for Kotlin-based code generation, since it would break compilation? Or rather, do we also want to add this null check to our `ProcessorUnit`s in Java?

shiraji:
I would like to add null check in Java since it should happens in Java, too.

hotchemi:
🙇

hotchemi:
@tomoya0x00 Just in case, are you working on this issue?

tomoya0x00:
@hotchemi Yes!

But I'm sorry.
Currently, I prioritize other tasks, and work on this issue will be resumed in latter half of next week.

hotchemi:
https://twitter.com/renaud_mathieu/status/1019222211004129282

tomoya0x00:
@hotchemi Thank you for your review!! ( #511 )

If you like, I'd create PR on your idea during the weekend.

hotchemi:
@tomoya0x00 Go ahead thank you!

hotchemi:
Will be released in next milestone!

PR #511, opened 2018-08-08
Add handling support fragment's getActivity()

tomoya0x00:
See #483


tomoya0x00:
🤔

hotchemi:
👀

hotchemi:
^ I think we don't need to check `rpe.isSubtypeOf` because there's no guarantee that activity is not null so just adding null check statement is ok, thx!

tomoya0x00:
OK. I will adding null check every time.

The generated code will be as below.
```java
static void showCameraWithPermissionCheck(MainActivity target) {
  Activity activity = target;
  if (activity == null) return;
  if (PermissionUtils.hasSelfPermissions(activity, PERMISSION_SHOWCAMERA)) {
    target.showCamera();
  } else {
    if (PermissionUtils.shouldShowRequestPermissionRationale(target, PERMISSION_SHOWCAMERA)) {
      target.showRationaleForCamera(new MainActivityShowCameraPermissionRequest(target));
    } else {
      ActivityCompat.requestPermissions(target, PERMISSION_SHOWCAMERA, REQUEST_SHOWCAMERA);
    }
  }
}
```

If checking that target inherits Activity is as below.
```java
static void showCameraWithPermissionCheck(MainActivity target) {
  if (target == null) return;
  if (PermissionUtils.hasSelfPermissions(target, PERMISSION_SHOWCAMERA)) {
    target.showCamera();
  } else {
    if (PermissionUtils.shouldShowRequestPermissionRationale(target, PERMISSION_SHOWCAMERA)) {
      target.showRationaleForCamera(new MainActivityShowCameraPermissionRequest(target));
    } else {
      ActivityCompat.requestPermissions(target, PERMISSION_SHOWCAMERA, REQUEST_SHOWCAMERA);
    }
  }
}
```

Which do you think is better?

hotchemi:
I think the latter is better! Thank you for your suggestion 😃

hotchemi:
Oh sorry for the late..can I ask the current status of this PR? 👀 @tomoya0x00

hotchemi:
@tomoya0x00 Sorry for the late and let me share my idea.

- I noticed we can solve this problem in most cases by adding `NonNull` annotation.

```java
static void showCameraWithPermissionCheck(@NonNull MyActivity target) {
}
```

- The only exceptional case is getting activity from fragment, but we expect that fragment is attached to an activity so we can use `requireActivity` method which is included in new support library.

```diff
- Activity activity = target.getActivity();
+ Activity activity = target.requireActivity();
```

It would cause `IllegalStateException` but I think it's better to notify developer that something wrong is happening, rather than do nothing.

So I wanna close this PR(really sorry for that). I'd create new PR but if you wanna continue working on it let us know 🙇 Thank you for your great work!

PR #526, opened 2018-09-17
Add NonNull annotation and use requireActivity

tomoya0x00:
See #483 and https://github.com/permissions-dispatcher/PermissionsDispatcher/pull/511#issuecomment-420142755


hotchemi:
Thank you! Let me check 👀

PR #529, opened 2018-09-22
Make RequestCodeProvider singleton

hotchemi:
To address https://github.com/permissions-dispatcher/PermissionsDispatcher/issues/519.


processor/src/main/kotlin/permissions/dispatcher/processor/PermissionsProcessor.kt
```diff
@@ -52,9 +52,7 @@ class PermissionsProcessor : AbstractProcessor() {
     }
 
     override fun process(annotations: Set<TypeElement>, roundEnv: RoundEnvironment): Boolean {
-        // Create a RequestCodeProvider which guarantees unique request codes for each permission request
-        val requestCodeProvider = RequestCodeProvider()
-
+        val requestCodeProvider = RequestCodeProvider
```
mannodermaus: Do we still need this val? If the `RequestCodeProvider` is accessible anyway, we don't have to pass it around anymore, I think

processor/src/test/java/permissions/dispatcher/processor/data/Source.java
```diff
@@ -763,7 +763,7 @@ protected String getName() {
                     "import java.lang.String;",
                     "import permissions.dispatcher.PermissionUtils;",
                     "final class MyActivityPermissionsDispatcher {",
-                    "   private static final int REQUEST_SHOWCAMERA = 0;",
+                    "   private static final int REQUEST_SHOWCAMERA = 12;",
```
mannodermaus: I don't think this is reliable enough for us, because the order of tests might change over time, and this will become a huge hassle for us. Instead, let's consider opening up the `currentCode` field (maybe using `@VisibleForTesting` to highlight the purpose), and resetting its value to 0 before each test

technoir42:
Have you considered how this would work if annotation processing was incremental (which would be great to implement btw)? Would `RequestCodeProvider` instance the same among AP invocations?

processor/src/main/kotlin/permissions/dispatcher/processor/PermissionsProcessor.kt
hotchemi: I just kept current interface deliberately since there's no guarantee to solve the issue. After confirming I'd like to fix more fundamentally.

processor/src/test/java/permissions/dispatcher/processor/data/Source.java
hotchemi: Umm yea..do you come up with any good idea? 🤔

hotchemi:
@technoir42 yea good point. Might be my fault actually 🤔

processor/src/main/kotlin/permissions/dispatcher/processor/PermissionsProcessor.kt
mannodermaus: OK!

hotchemi:
Umm I think this is not an optimal way to address the issue. Let me close

processor/src/test/java/permissions/dispatcher/processor/data/Source.java
mannodermaus: Opening `RequestCodeProvider`'s field should be enough:

```kotlin
object RequestCodeProvider {

    @VisibleForTesting
    val currentCode = AtomicInteger(0)

    fun nextRequestCode(): Int = currentCode.andIncrement
}
```

Then, inside our test class:

```kotlin
@Before
fun beforeEach() {
    RequestCodeProvider.currentCode = 0
}
```

hotchemi:
Continue discussion at https://github.com/permissions-dispatcher/PermissionsDispatcher/issues/519#issuecomment-424313966

PR #536, opened 2018-10-28
Fix CallNeedsPermissionDetector to scan only annotated classes

hotchemi:
Fixes https://github.com/permissions-dispatcher/PermissionsDispatcher/issues/502


shiraji:
It seems to me too restricted to requires signed commits

PR #564, opened 2018-12-17
Improve performance by memoization technique

hotchemi:


hotchemi:
Not so much improvement than expected..let me close 🤔

PR #571, opened 2018-12-21
Prepare for 4.2.0

hotchemi:


hotchemi:
We're waiting response from bintray support.

PR #573, opened 2018-12-21
Add a link back to the repo from the website

TWiStErRob:


TWiStErRob:
I put it above because the first 3 now are related to the repo (devs need to know these), and the last one is related to third parties (devs might want to know this).

TWiStErRob:
Happy to swap though if you want, because I see that now there's an external link in the middle of the doc links.

hotchemi:
Let me merge and consider that there's a better way to show repo link!

PR #606, opened 2019-04-05
Add support for internal classes

### LINKED ISSUE #605: Support internal fragment definition
Requested by user.
https://gist.github.com/satoshun/51e73ff509363dcdf7d463695d2fe6f1

mannodermaus:
Side effect: This PR raises the Kotlin requirement to 1.3 or newer. Since then, the `@kotlin.Metadata` annotation has been accessible through the standard library. With the help of a companion library, `kotlinx-metadata-jvm`, we can inspect the flags of an annotated class and see if it's `internal` or not. In case it is, the generated permission methods for Kotlin will also have the same modifier applied to them.

So far, I verified this change through a manual test. Where would we put a unit test for this? There don't seem to be any tests against `androidx.fragment.app.Fragment` in the codebase anymore (I haven't checked out that part in a while, haha).

Resolves #605.


shiraji:
> There don't seem to be any tests against androidx.fragment.app.Fragment in the codebase anymore

Is this the one?

https://github.com/permissions-dispatcher/PermissionsDispatcher/blob/master/processor/src/test/java/permissions/dispatcher/processor/data/Source.java#L758

You can also add test cases to [test module](https://github.com/permissions-dispatcher/PermissionsDispatcher/tree/master/test/src/test/java/permissions/dispatcher/test)

mannodermaus:
`Source.java` only has Java-based tests. I added a test to `KtProcessorTestSuite.kt` now!

PR #608, opened 2019-04-12
Fix Conductor tests

technoir42:
Fixes:
```
org.mockito.exceptions.misusing.InvalidUseOfMatchersException: 
Misplaced argument matcher detected here:

-> at permissions.dispatcher.test.ExtensionsKt.mockShouldShowRequestPermissionRationaleConductorController(Extensions.kt:83)

You cannot use argument matchers outside of verification or stubbing.
```


hotchemi:
@technoir42 Hi I'm really sorry for being late🙇Can I ask why this change is needed? I thought we don't need this since CI basically passes on TravisCI? let me kn ow if I'm missing 🙇

technoir42:
These tests fail locally when I run `./gradlew clean test` on master and also fail from IDE. It's weird that they pass on TravisCI.

Can you try running them locally in both ways?

PR #625, opened 2019-06-10
Update Gradle, AGP, Kotlin and KotlinPoet

technoir42:
Bumping some versions before incremental apt. We'll need KotlinPoet 1.3.0 for sure which requires the most of these changes.
I decided not to update Gradle to 5.x for now because it might break artifact uploads.


test/build.gradle
```diff
@@ -4,7 +4,6 @@ apply plugin: 'kotlin-kapt'
 
 android {
     compileSdkVersion COMPILE_SDK_VERSION
-    buildToolsVersion BUILD_TOOLS_VERSION
```
hotchemi: Don't we need it?

technoir42: Setting `buildToolsVersion` has been optional since 3.0.0. I removed it so we don't have to update it manually anymore.
I can revert this if you prefer to set `buildToolsVersion` explicitly.

hotchemi: OK! I'm fine for the change.

PR #626, opened 2019-06-10
Incremental annotation processing for Kotlin

technoir42:
Depends on #625, https://github.com/permissions-dispatcher/kompile-testing/pull/13
Will add some tests later.

### DIFF
--- a/processor/src/main/kotlin/permissions/dispatcher/processor/PermissionsProcessor.kt
+++ b/processor/src/main/kotlin/permissions/dispatcher/processor/PermissionsProcessor.kt
@@ -4,7 +4,6 @@ import permissions.dispatcher.RuntimePermissions
 import permissions.dispatcher.processor.impl.javaProcessorUnits
 import permissions.dispatcher.processor.impl.kotlinProcessorUnits
 import permissions.dispatcher.processor.util.findAndValidateProcessorUnit
-import java.io.File
 import javax.annotation.processing.AbstractProcessor
 import javax.annotation.processing.Filer
 import javax.annotation.processing.ProcessingEnvironment
@@ -14,7 +13,6 @@ import javax.lang.model.element.Element
 import javax.lang.model.element.TypeElement
 import javax.lang.model.util.Elements
 import javax.lang.model.util.Types
-import javax.tools.Diagnostic
 import kotlin.properties.Delegates
 
 /** Element Utilities, obtained from the processing environment */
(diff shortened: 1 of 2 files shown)


hotchemi:
👀

gradle.properties
```diff
@@ -30,7 +30,7 @@ LINT_VERSION            = 26.3.2
 ROBOLECTRIC_VERSION     = 3.3.2
 COMMONS_IO_VERSION      = 2.6
 CONDUCTOR_VERSION       = 2.1.5
-KOMPILE_TESTING_VERSION = 0.1.1
+KOMPILE_TESTING_VERSION = 0.1.3
```
hotchemi: Can you rebase?👀https://github.com/permissions-dispatcher/PermissionsDispatcher/pull/628

processor/src/main/kotlin/permissions/dispatcher/processor/impl/kotlin/KotlinBaseProcessorUnit.kt
```diff
@@ -109,6 +109,7 @@ abstract class KotlinBaseProcessorUnit : KtProcessorUnit {
 
     private fun createWithPermissionCheckFun(rpe: RuntimePermissionsElement, method: ExecutableElement): FunSpec {
         val builder = FunSpec.builder(withPermissionCheckMethodName(method))
+                .addOriginatingElement(rpe.element)
```
hotchemi: Do we need to add an original element for every spec builder?👀

processor/src/main/kotlin/permissions/dispatcher/processor/PermissionsProcessor.kt
```diff
@@ -68,21 +61,9 @@ class PermissionsProcessor : AbstractProcessor() {
     }
 
     private fun processKotlin(element: Element, rpe: RuntimePermissionsElement, requestCodeProvider: RequestCodeProvider) {
-        // FIXME: weirdly under kaptKotlin files is not recognized as source file on AS or IntelliJ
-        // so as a workaround we generate .kt file in generated/source/kapt/$sourceSetName
-        // ref: https://github.com/hotchemi/PermissionsDispatcher/issues/320#issuecomment-316175775
-        val kaptGeneratedDirPath = processingEnv.options[KAPT_KOTLIN_GENERATED_OPTION_NAME]?.replace("kaptKotlin", "kapt")
```
hotchemi: Not sure the issue has been already resolved or not👀 
https://youtrack.jetbrains.com/issue/KT-20269

technoir42: Yes, that was fixed.

processor/src/main/kotlin/permissions/dispatcher/processor/impl/kotlin/KotlinBaseProcessorUnit.kt
technoir42: It might work if we set an origin element for just one spec that's guaranteed to be present in the file. But I think it's better to be safe and explicit and set origin element for every top-level spec we produce.
KotlinPoet 1.3.0+ deduplicates origin elements before passing them to Filer so kapt will see only one origin element.

PR #627, opened 2019-06-11
602:Improve method identification logic

hotchemi:
## Issues

- address https://github.com/permissions-dispatcher/PermissionsDispatcher/issues/602

## Overview

- Fix to use parent class name and method name info to identify method.


hotchemi:
Let me merge.

Issue #642, opened 2019-09-25
Generated kotlin files still use Java collections


kanat:
## Overview

I use kotlin in combination with kapt.

Suppose you've got some function below:
```kotlin
class SomeActivity : AppCompatActivity() {
    @NeedsPermission(CAMERA, RECORD_AUDIO)
    fun doWork(values: kotlin.collections.Set<kotlin.String>) {
        ...
    }
}
```

The `SomeActivityPermissionsDispatcher.kt` will be generated, but with different input type in the function:
```kotlin
fun SomeActivity.doWorkWithPermissionCheck(values: java.util.Set<java.lang.String>) {
    ...
}
```

### Expected

- `kotlin.collections.*` collections are used in generated files

### Actual

- `java.util.*` collections are used in generated files

## Environment

`implementation 'org.permissionsdispatcher:permissionsdispatcher:4.5.0'`
`kapt 'org.permissionsdispatcher:permissionsdispatcher-processor:4.5.0'`

hotchemi:
thank you! this originates from kotlinpoet seemingly and we have a workaround method to address similar issue. Would you mind send a PR?

https://github.com/permissions-dispatcher/PermissionsDispatcher/blob/f9175b1ec883c054750783e71d1992a29bd05a49/processor/src/main/kotlin/permissions/dispatcher/processor/util/Extensions.kt#L122

kanat:
> thank you! this originates from kotlinpoet seemingly and we have a workaround method to address similar issue. Would you mind send a PR?
> 
> https://github.com/permissions-dispatcher/PermissionsDispatcher/blob/f9175b1ec883c054750783e71d1992a29bd05a49/processor/src/main/kotlin/permissions/dispatcher/processor/util/Extensions.kt#L122

Thanks for navigating me to problematic point. Please, check the PR.

PR #643, opened 2019-09-25
fix Java to Kotlin collections mapping (#642)

kanat:


hotchemi:
thx! in review

processor/src/main/kotlin/permissions/dispatcher/processor/util/Extensions.kt
```diff
@@ -122,15 +122,19 @@ fun FileSpec.Builder.addTypes(types: List<TypeSpec>): FileSpec.Builder {
 fun TypeName.correctJavaTypeToKotlinType(): TypeName {
     return if (this is ParameterizedTypeName) {
         val typeArguments = this.typeArguments.map { it.correctJavaTypeToKotlinType() }.toTypedArray()
-        return this.rawType.parameterizedBy(*typeArguments)
+        val rawType = ClassName.bestGuess(this.rawType.correctJavaTypeToKotlinType().toString())
```
hotchemi: why do we need this line change? 👀

processor/src/main/kotlin/permissions/dispatcher/processor/util/Extensions.kt
```diff
@@ -122,15 +122,19 @@ fun FileSpec.Builder.addTypes(types: List<TypeSpec>): FileSpec.Builder {
...
+        return rawType.parameterizedBy(*typeArguments)
     } else when (toString()) {
         "java.lang.Byte" -> ClassName("kotlin", "Byte")
         "java.lang.Double" -> ClassName("kotlin", "Double")
         "java.lang.Object" -> ClassName("kotlin", "Any")
         "java.lang.String" -> ClassName("kotlin", "String")
+        "java.util.Set" -> ClassName("kotlin.collections", "MutableSet")
```
hotchemi: not `kotlin.collections.Set`? 👀

processor/src/main/kotlin/permissions/dispatcher/processor/util/Extensions.kt
```diff
@@ -122,15 +122,19 @@ fun FileSpec.Builder.addTypes(types: List<TypeSpec>): FileSpec.Builder {
...
+        "java.util.List" -> ClassName("kotlin.collections", "MutableList")
```
hotchemi: ditto

processor/src/main/kotlin/permissions/dispatcher/processor/util/Extensions.kt
kanat: Because collection is a parametrized type.
Without that change transition will look like:
`java.util.Set<java.lang.String>` -> `java.util.Set<kotlin.String>`.

It's not enough just to convert `typeArguments`, we have to convert collection itself (`rawType`).

processor/src/main/kotlin/permissions/dispatcher/processor/util/Extensions.kt
kanat: `java.util.Set` is more like `kotlin.collections.MutableSet`.

If you declare method like
```kotlin
@NeedsPermission(CAMERA, RECORD_AUDIO)
fun doWork(values: kotlin.collections.MutableSet<kotlin.String>) {
    ...
}
```
The generated code won't be compilable, because it won't be able to pass `kotlin.collections.Set` into method with `kotlin.collections.MutableSet` parameter.
And vice versa `kotlin.collections.MutableSet` fits `kotlin.collections.Set`.

PR #644, opened 2019-10-22
Remove conductor support

hotchemi:
### Issue

ref: https://github.com/permissions-dispatcher/PermissionsDispatcher/issues/595

### Overview

After supporting Conductor some of the users have reported compilation error issue above.
I've investigated a bit but it's quite uncertain how it occurs(perhaps not our bug...).
That's one thing and in terms of the extensibility of the library, just supports conductor is not effective enough and I'd like to look for another approach.


Sar777:
hi, do you have plan to support Conductor library? maybe in the other module? thanks!

hotchemi:
@Sar777 thank you! I noticed nowadays we have a bunch of similar concept libraries and don't have time to support every libs. So I'd like to consider a more flexible way to support those things. Let us know if you have a good one🙇

PR #652, opened 2019-12-10
Add missing dangerous permissions in later versions

hotchemi:


library/src/main/java/permissions/dispatcher/PermissionUtils.java
```diff
@@ -13,18 +13,25 @@
 public final class PermissionUtils {
     // Map of dangerous permissions introduced in later framework versions.
     // Used to conditionally bypass permission-hold checks on older devices.
+    // ref: https://developer.android.com/reference/android/Manifest.permission
     private static final SimpleArrayMap<String, Integer> MIN_SDK_PERMISSIONS;
 
     static {
-        MIN_SDK_PERMISSIONS = new SimpleArrayMap<String, Integer>(8);
+        MIN_SDK_PERMISSIONS = new SimpleArrayMap<String, Integer>(14);
+        MIN_SDK_PERMISSIONS.put("android.permission.USE_SIP", 9);
```
mannodermaus: 👀

shiraji: 👀

hotchemi: hehe

zloyreznic:
Where I can find snapshot version?
Here is nothing http://oss.jfrog.org/artifactory/oss-snapshot-local/org/permissionsdispatcher/permissionsdispatcher/4.7.0/permissionsdispatcher-4.7.0.pom

hotchemi:
@zloyreznic we've noticed recently we haven't released snapshot for a while since changing groupId and now requesting access to jfrog again. Thank you for the heads up.
Btw do you need this change specifically?

PR #657, opened 2020-02-04
Official Gradle Wrapper Validation Action

JLLeitschuh:
See: https://github.com/gradle/wrapper-validation-action


hotchemi:
Thx! Since we're not familiar with the action let us take a look 🙇

PR #688, opened 2020-09-10
Solve multiple call of events

mtrakal:
https://github.com/permissions-dispatcher/PermissionsDispatcher/issues/687


hotchemi:
Thank you, I was reading https://medium.com/androiddevelopers/livedata-with-snackbar-navigation-and-other-events-the-singleliveevent-case-ac2622673150 and it said using `Event` like the following is more recommended as `SingleLiveEvent is that it’s restricted to one observer.`, what do you think?


```kotlin
/**
 * Used as a wrapper for data that is exposed via a LiveData that represents an event.
 */
open class Event<out T>(private val content: T) {

    var hasBeenHandled = false
        private set // Allow external read but not write

    /**
     * Returns the content and prevents its use again.
     */
    fun getContentIfNotHandled(): T? {
        return if (hasBeenHandled) {
            null
        } else {
            hasBeenHandled = true
            content
        }
    }

    /**
     * Returns the content, even if it's already been handled.
     */
    fun peekContent(): T = content
}
```

hotchemi:
https://github.com/permissions-dispatcher/PermissionsDispatcher/pull/689

hotchemi:
Let me close the PR since we've already merged https://github.com/permissions-dispatcher/PermissionsDispatcher/pull/689, thank you for the support!

PR #711, opened 2021-03-02
fix: use Either to handle both Activity and Fragment with the same interface

hotchemi:


ktx/src/main/java/permissions/dispatcher/ktx/PermissionsRequesterImpl.kt
```diff
@@ -1,39 +1,81 @@
...
+        is FragmentType -> right.requireContext()
+    }
+
+private val ViewType.viewModelStoreOwner: ViewModelStoreOwner
+    get() = when (this) {
+        is ActivityType -> left
+        is FragmentType -> right
+    }
+
+private val ViewType.lifeCycleOwner: LifecycleOwner
```
mannodermaus, line 71:
```suggestion
private val ViewType.lifecycleOwner: LifecycleOwner
```

For consistency with the AndroidX property ☀️

sample/build.gradle
```diff
@@ -18,6 +18,9 @@ android {
     sourceSets {
         main.java.srcDirs += "src/main/kotlin"
     }
+    lintOptions {
+        abortOnError false
```
hotchemi, line 22: somehow the error occurs only on GitHub Action, not on my local 🤔 

https://github.com/permissions-dispatcher/PermissionsDispatcher/runs/2008862691?check_suite_focus=true#step:6:569

ktx/src/main/java/permissions/dispatcher/ktx/PermissionsRequesterImpl.kt
hotchemi: thx bro! I'll think something over entirely and polish PR with description😄

hotchemi:
just noticed another solution. let me close the PR

PR #742, opened 2021-08-12
fix: keep a reference not to release requestFun in KtxPermissionRequest

hotchemi:
resolves https://github.com/permissions-dispatcher/PermissionsDispatcher/issues/734


SamYStudiO (approved):
This works indeed!

PR #757, opened 2021-11-23
fix: use ContextCompat instead of PermissionChecker

hotchemi:
ref: https://github.com/permissions-dispatcher/PermissionsDispatcher/issues/756


mannodermaus (approved):
Very curious why the AndroidX method would raise the privacy indicator. Almost sounds like a bug in androidx.core. 🤔

hotchemi:
obviously. BTW I have noticed it's kinda hard to maintain test module...we shall move on to E2E test or like that!

atkunm (approved):

PR #772, opened 2022-06-16
Fix for issue 767 in forked repo

### LINKED ISSUE #767: [ktx] onNeverAskAgain not called
## FAQs

 
___

## Overview

onNeverAskAgain callback is not called when "Don't ask again" is selected.

### Expected

onNeverAskAgain callback is called, so I can show the appropriate UI.

### Actual

onNeverAskAgain callback is not called.

## Environment

- Running ktx-sample on API 29 and API 30

## Investigation
After debugging it, I can see that `onNeverAskAgain: WeakReference<Fun>?` is null in `PermissionRequestViewModel.invoke()`.
After reading through the closed issues, I came across https://github.com/permissions-dispatcher/PermissionsDispatcher/issues/735. Wondering the weakreference could be the issue? 

## Reproducible steps

https://user-images.githubusercontent.com/35034828/158357946-129980bf-6547-4228-82a4-a03399f88269.mp4

francescopedronomnys:
Keep a strong reference to onNeverAskAgain to avoid the WeakRefence to it stored on PermissionRequestViewModel to become null when user denies permission multiple times.
This should resolve #767 .

### DIFF
--- a/ktx/src/main/java/permissions/dispatcher/ktx/PermissionsRequesterImpl.kt
+++ b/ktx/src/main/java/permissions/dispatcher/ktx/PermissionsRequesterImpl.kt
@@ -11,7 +11,7 @@ internal class PermissionsRequesterImpl(
     private val onShowRationale: ShowRationaleFun?,
     private val onPermissionDenied: Fun?,
     private val requiresPermission: Fun,
-    onNeverAskAgain: Fun?,
+    private val onNeverAskAgain: Fun?,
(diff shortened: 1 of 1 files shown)


dkornev (approved):
LGTM

I see no possible issues with this fix. And since we could reproduce #767 100% time after the GC run, it would be good to merge this one.

vladyslav-dovhal-quarks-tech (approved):

amansinghrajpoot (approved):

PR #782, opened 2023-11-10
fix 779 issus

shaomaicheng:
Reason for the change:

Based on the issue: https://github.com/permissions-dispatcher/PermissionsDispatcher/issues/779

Bug reason:

When upgrading Kotlin or base library versions, some libraries that use KotlinPoet may also be upgraded to version 1.11 or higher.

KotlinPoet 1.11.0 has the following release note:

• Fix: Enforce only allowed parameter modifiers in ParameterSpec (i.e. crossinline, vararg, and noinline).

For more details, please refer to: https://github.com/square/kotlinpoet/releases

When generating the PermissionRequest class for methods annotated with @NeedsPermission, the private constructor will cause an error.

Reproduction steps:

Upgrade KotlinPoet to 1.11.0
Add a new parameter "callback" to the MainActivity#showCamera method in the sample Modification:

Do not pass the KModifier.*PRIVATE parameter


shaomaicheng:
@hotchemi

